PR #5, opened 2019-12-02
Retrieve public keys for multiple addresses

rsbondi:
This will enable a user to verify and spend from multiple addresses in a tool such as caravan.

For public key verification, a public key will be returned as a string for a single public key represented by a single BIP32 path maintaining backward compatibility.  For multiple addresses, referring to multiple BIP32 paths, an array of public keys is returned.

An extra named argument `bip32Paths` has been added to the exports which will take an array of BIP32 paths.  It is required to provide either the single `bip32Path` argument or the multiple `bip32Paths` argument.

Trezor provides a separate API call for multiple BIP32 paths.  Ledger's API does not, each public key for each path requires a separate API call, and therefore a confirmation on the device for each key that is exported.  These are done in a loop and stored and returned as an array, similar to how Trezor naturally does it.  The Trezor code loops for message validation looping through all provided BIP32 paths

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,12 +6,12 @@ export const TREZOR = 'trezor';
 export const LEDGER = 'ledger';
 export const HERMIT = 'hermit';
 
-export function HardwareWalletExportPublicKey({walletType, network, bip32Path}) {
+export function HardwareWalletExportPublicKey({walletType, network, bip32Path, bip32Paths}) {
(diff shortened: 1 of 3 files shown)


waldenraines:
@dhruvbansal @howech is this still change still desired?

waldenraines:
Closing due to inactivity.

PR #9, opened 2020-02-20
Add ledger photos

kajalkris:
First part of Ledger photos added to exporting public key + what looks like Dhruv's code.


dhruvbansal:
These images and the corresponding code was merged in via https://github.com/unchained-capital/unchained-wallets/pull/17

PR #11, opened 2020-02-29
Ledger xpub support

rsbondi:
This requires the retrieval of the parent public key so the node being exported will have the proper parent `parentFingerprint`.  An additional call is made to get the parent but skipping the verify since this is done in the call to `super`

See [this issue](https://github.com/LedgerHQ/ledgerjs/issues/156#issuecomment-422184943) for some context.  The call to retrieve the parent returns with the fingerprint, so no need to calculate.

This supercedes #4 and is merged with #9


dhruvbansal:
This feature was added via https://github.com/unchained-capital/unchained-wallets/pull/17

PR #12, opened 2020-02-29
correct hermit BIP32 path from QR code

rsbondi:
Hermit encodes with `bip32_path` and not `bip32Path`


waldenraines:
@dhruvbansal @howech is this still change still desired?

waldenraines:
Closing due to inactivity.

PR #14, opened 2020-03-04
docs: remove from master and publish directly to gh-pages

bucko13:
This PR removes docs/* from the main branch and includes a script that builds docs and deploys to dedicated gh-pages branch.


bucko13:
@dhruvbansal this should be good to go for getting rid of docs from master. Without write access I can't confirm the settings for reading the github pages site, but that probably needs to be updated to draw from `gh-pages` branch rather than `docs/` on `master`. 

Also, I added a new deployment for docs building when on master branch. @howech not sure if you have any opinions on this, but let me know if you think there would be a better way/frequency. 

Once this is approved, I'll apply the same changes to unchained-bitcoin (which I think is the only other one that needs it)

PR #15, opened 2020-03-05
build-docs: fix script bugs

bucko13:


howech:
Please move the script to a bin directory and update the .npmignore file.

PR #18, opened 2020-03-19
check env for trezor dev

bucko13:
When testing updates to trezor connect, wallets needs a way to know to connect to the local instance of connect rather than the live one at https://connect.trezor.io/8. 

I know there was some questions around the best way to set configs for trezor connect and that there was some hesitancy to do any dependency injection. This probably isn't the ideal way to go about what's needed, but this instance of wanting to set a custom `connectSrc`, particularly for development seems useful, and better than simply replacing over the code when it is needed. Another option could be setting something like `TREZOR_HOST` at runtime too instead of the very narrow `TREZOR_DEV` that I have here. I'm open to suggestions, but wanted to kick off the discussion!

For quick background, this code was necessary for testing the P2WSH bug and patch (https://github.com/trezor/connect/pull/541)


waldenraines:
@bucko13 is this still a desired change?

humanumbrella (commented):
Not sure about the changes in package-lock?

I'm guessing the default port for running your own TC is on 8088 and that running TC.init isn't necessary *unless* you want to specify a different `connectSrc` ? 

Pretty sure I answered my own questions, but just to confirm:  

https://wiki.trezor.io/Developers_guide:Running_Trezor_Connect_on_localhost

and

https://github.com/trezor/connect/blob/develop/docs/index.md#running-local-version-developstable

Once clarity on packages changes I can approve - this seems like a fine solution for the problem of setting things up locally.

humanumbrella (changes requested):
Can you add a section to the README highlighting this change so others don't have to dig in to figure out how to do this?

humanumbrella (approved):
LGTM

PR #22, opened 2020-05-30
Updates formatting and fix bugs

humanumbrella:
This PR mostly contains style changes and automatic fixes. However, there were a few bugs and linted errors. E.g. this PR changes several `return await` in async functions to `return` per https://eslint.org/docs/rules/no-return-await and finally fixes a couple additional typos.

Tests passing.

### DIFF
--- a/src/ledger.js
+++ b/src/ledger.js
@@ -1,5 +1,3 @@
-/* eslint-disable max-lines */
-
 /**
  * This module provides classes for Ledger hardware wallets.
  *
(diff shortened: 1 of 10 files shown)


src/ledger.js
```diff
@@ -183,9 +195,9 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
    * }
    */
   async withApp(callback) {
```
bucko13, line 197: If the awaits are removed, does this still need to be async?

humanumbrella: Yes - I believe you're right here. I removed the async in the new PR I submitted. (closing this one). 

https://github.com/unchained-capital/unchained-wallets/pull/24/files#diff-5da93726c6b0fe6dfa4d1b76cdc3767dL185 but you have to open the diff on ledger.js

PR #24, opened 2020-06-22
webusb and no verify xpub export

humanumbrella:
This PR mostly contains style changes and automatic fixes. However, there were a few bugs and linted errors. E.g. this PR changes several return await in async functions to return per https://eslint.org/docs/rules/no-return-await and finally fixes a couple additional typos.

Tests passing, and test suite ran locally with success.


humanumbrella:
`Ledger Nano S v1.6.0 (MCU v1.12)`

`Trezor Model 1 v.1.8.1`

Screen Shot 2020-06-22 at 11 38 58 AM
Screen Shot 2020-06-22 at 11 46 35 AM

src/hermit.js
```diff
@@ -212,15 +210,15 @@ export class HermitExportPublicKey extends HermitReader {
       throw new Error("No BIP32 path in QR code.");
     }
     result.bip32Path = bip32Path;
-    delete result.bip32_path;
+    Reflect.deleteProperty(result, "bip32_path");
```
bucko13, line 213: wow, I've never heard of `Reflect` before! Neat.

bucko13: I know this is just a lint cleanup, but what exactly is going on here? Do you know why there's this rigamarole with two different `bip32Path` properties?

src/hermit.js
```diff
@@ -345,10 +343,10 @@ export class HermitSignTransaction extends HermitDisplayer {
 
     if (!this.inputsAreSupported()) {
       messages.push({
-        statet: UNSUPPORTED,
+        state: UNSUPPORTED,
```
bucko13, line 346: Wow, good catch.

src/ledger.js
```diff
@@ -159,8 +172,9 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
    * }
    */
   async withTransport(callback) {
-    const transport = await TransportU2F.create();
-    return await callback(transport);
+    const transport = await TransportWebUsb.create();
+    // const transport = await TransportU2F.create();
```
bucko13, line 176: nit: cleanup

src/ledger.js
```diff
@@ -589,15 +645,15 @@ class LedgerExportHDNode extends LedgerBitcoinInteraction {
    * @returns {object} the HD node object.
    */
   async run() {
```
bucko13, line 647: Don't think we need either of the `async`s here right?

src/ledger.js
```diff
@@ -682,18 +739,18 @@ export class LedgerExportExtendedPublicKey extends LedgerExportHDNode {
...
-    const pubkey = await this.getParentPublicKey()
-    return getFingerprintFromPublicKey(pubkey)
-   }
-  
+  async getFingerprint() {
+    const pubkey = await this.getParentPublicKey();
+    return getFingerprintFromPublicKey(pubkey);
+  }
+
   async getParentPublicKey() {
```
bucko13, line 747: Don't need this one either anymore.

src/ledger.js
```diff
@@ -911,11 +968,11 @@ export class LedgerSignMultisigTransaction extends LedgerBitcoinInteraction {
    * Input signatures produced will always have a trailing `...01`
    * {@link https://bitcoin.org/en/glossary/sighash-all SIGHASH_ALL}
    * byte.
-   * 
+   *
    * @returns {string[]} array of input signatures, one per input
    */
   async run() {
```
bucko13, line 974: Same.

src/ledger.js
```diff
@@ -53,10 +51,10 @@ import IMAGES from "./images";
 export const LEDGER = 'ledger';
 
 const bitcoin = require('bitcoinjs-lib');
-const bip32 = require('bip32');
 
-const TransportU2F = require("@ledgerhq/hw-transport-u2f").default;
-const LedgerBtc    = require("@ledgerhq/hw-app-btc").default;
+// const TransportU2F = require("@ledgerhq/hw-transport-u2f").default;
```
bucko13, line 55: nit: cleanup

bucko13:
I think you may need to run `npm install` and commit the changes to the `package-lock.json` file in order to make travis happy: 
https://travis-ci.com/github/unchained-capital/unchained-wallets/builds/172541993

src/hermit.js
humanumbrella: > wow, I've never heard of `Reflect` before! Neat.
Same -- this was an IDE mediated swap.

I'm not sure about the rigamarole, except looking at `parseHermitQRCodeData` looks like it grabs the data and tries to export JSON. To me this means that hermit exports labeled JSON into the QR code and the labels don't match the shape of the library (or other keystores). This looks like where that issue is corrected.

src/ledger.js
humanumbrella: Aiight, I'll also remove the package

src/ledger.js
humanumbrella: there's an await in the return

src/ledger.js
humanumbrella: this one needs bc await call to signP2SHTransaction

src/ledger.js
bucko13: So that should be covered by the async preceding the callback function. Since we're not awaiting anything in this scope, then it should be fine. 

It would be the same as if you wrote it like this:

```js
getParentPublicKey() {
  // just declaring the function here
  async function callbackFn(app) {
        const parentPath = getParentPath(this.bip32Path)	      
        return (await app.getWalletPublicKey(parentPath)).publicKey;
  }

  return this.withApp(callbackFn);
}
```

src/hermit.js
bucko13: Oh yeah, you're right. I think i remember Destry patching this previously to fix a bug :-/

src/ledger.js
humanumbrella: This makes sense to me but let's chat tomorrow

humanumbrella:
forgot to respond to the other commentary, but the `no verify xpub export` part
is located at https://github.com/unchained-capital/unchained-wallets/pull/24/files#diff-5da93726c6b0fe6dfa4d1b76cdc3767dL593 (open diff on `ledger.js` for this link to do any highlighting properly)

.eslintrc.json
```diff
@@ -108,7 +108,7 @@
         "max-classes-per-file": "off",
         "max-depth": "error",
         "max-len": "off",
-        "max-lines": "error",
+        "max-lines": "off",
```
waldenraines, line 111: What is the rationale for this?

humanumbrella: Most of the files in this package are over 300 lines (arbitrary) so this causes errors that aren't actually errors. I'm open to suggestion -- this was easier than saying ignore this particular rule in each file.

humanumbrella:
oops, one sec - got trigger happy on review requests.

humanumbrella:
@waldenraines - you should be able to test now. NOTE: while you *can* use the Bitcoin Testnet app on the ledger for the webusb transport - you still cannot sign with the u2f transport (in firefox) without switching to the main Bitcoin app.

waldenraines:
> @waldenraines - you should be able to test now. NOTE: while you _can_ use the Bitcoin Testnet app on the ledger for the webusb transport - you still cannot sign with the u2f transport (in firefox) without switching to the main Bitcoin app.

Nice, will have a look.

waldenraines:
@bucko13 did you have any further comments?  Looks like you're still marked as "requesting changes".

src/ledger.js
```diff
@@ -40,7 +40,7 @@ import {
   DirectKeystoreInteraction,
 } from "./interaction";
 
-import IMAGES from "./images";
+//import IMAGES from "./images";
```
bucko13, line 43: Wow, we're not using the images at all anymore? I guess we don't use them or have them yet for signing?

kajalkris:
Here are the results of my testing. Tested with:

ledger nano s: bitcoin app 1.4.2, secure element 1.6.0
ledger nano x: bitcoin app 1.4.2, secure element 1.2.4-1

- For Ledger S on Mac & Windows, I can sign on Chrome & Brave browsers. I was not able to sign on Firefox. 
- For Ledger X on Mac, I was able to sign on Chrome, Firefox, and Brave.
- For Ledger X on Windows, I was able to sign on Firefox, but on Chrome & Brave, I was unable to see the option to connect to the Ledger X device. See screenshot below.
Screen Shot 2020-06-24 at 9 54 29 AM

src/ledger.js
humanumbrella: we were only showing images for xpubs/pubkeys at the moment

humanumbrella:
re: @kajalkris  looks like others have reported nano X not showing up on windows ... https://github.com/LedgerHQ/ledgerjs/issues/456 -- I don't know how to handle that one.

For Ledger S -- I could sign on FF on mac (u2f) but had to try multiple times sometimes to finally get it to succeed (it failed, but I just kept trying, I exited and re-launched the btc app)

src/ledger.js
waldenraines: > we were only showing images for xpubs/pubkeys at the moment

See also https://github.com/unchained-capital/caravan/issues/112

bucko13:
LGTM! Tested in Chrome and Safari. Safari also throws the transport error but this isn't something we claim to support either right?

waldenraines:
I think we should go ahead and merge this and if there is anything we can do on our side to fix ledger nano x then we can do that in a follow-up PR. Thoughts @bucko13 @humanumbrella?

PR #26, opened 2020-07-07
Consolidates top level Ledger/Trezor messages

humanumbrella:
Particularly for the ledger messages, would like to consolidate three messages into two.

### DIFF
--- a/src/trezor.js
+++ b/src/trezor.js
@@ -228,7 +228,7 @@ export class TrezorInteraction extends DirectKeystoreInteraction {
     messages.push({
       state: ACTIVE,
       level: INFO,
-      text: "Your browser should open a new Trezor Connect window.  If you do not see this window, ensure you have enabled popups for this site.",
+      text: "Your browser should now open a new window to Trezor Connect. Ensure you have enabled popups for this site.",
       code: "trezor.connect.generic",
     });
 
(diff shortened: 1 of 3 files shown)


src/trezor.js
```diff
@@ -228,7 +228,7 @@ export class TrezorInteraction extends DirectKeystoreInteraction {
     messages.push({
       state: ACTIVE,
       level: INFO,
-      text: "Your browser should open a new Trezor Connect window.  If you do not see this window, ensure you have enabled popups for this site.",
+      text: "Your browser should now open a new window to Trezor Connect. Ensure you have enabled popups for this site.",
```
waldenraines, line 231: For some reason "window to Trezor Connect" sounds odd to me.  Additionally, whose browser opens a window instead of a tab these days?

Maybe "Your browser should now open a new Trezor Connect tab..."?

bucko13:
Yeah, no real objections for me. I'm indifferent to window vs. tab. The only argument I'd think of to keep window is that I feel like sometimes window is used as a stand-in to mean either or. But it doesn't really matter ultimately. Reducing the number of total messages seems like a good idea though. Too many messages and people just end up ignoring the whole thing.

src/trezor.js
waldenraines: I'm also fine merging this as-is and refining the messaging later if need be.

humanumbrella: happy to change it to get it in, but I also think we will refine more later.

humanumbrella: Ah, just saw buck's comment - I think it's fine to leave it as is for now. Maybe the new designer will push us in that direction anyway.

I was trying to drop the length of the message by rewording it.

PR #28, opened 2020-07-15
Adds better error handling, refactor outputScript, add named params

humanumbrella:
Screen Shot 2020-07-14 at 10 02 27 PM

Saw above while adding error messages, refactored a bit to change the method signature and also to remove duplicated bitcoinjs code and choosing rather to import from `unchained-bitcoin` and rebuild the unsigned transaction that way.

npm tests pass, I also ran through the device test suite with ledger nano S

### DIFF
--- a/src/ledger.js
+++ b/src/ledger.js
@@ -14,7 +14,7 @@
  *
  * @module ledger
  */
-import BigNumber from "bignumber.js";
+
 import {
   bip32PathToSequence,
   hardenedBIP32Index,
@@ -22,7 +22,6 @@ import {
   scriptToHex,
   multisigRedeemScript,
   multisigWitnessScript,
-  TESTNET,
   P2SH,
   P2SH_P2WSH,
   P2WSH,
@@ -30,6 +29,7 @@ import {
   getParentPath,
   getFingerprintFromPublicKey,
   deriveExtendedPublicKey,
+  unsignedMultisigTransaction,
 } from "unchained-bitcoin";
 
 import {
@@ -40,6 +40,9 @@ import {
   DirectKeystoreInteraction,
 } from "./interaction";
 
+import {splitTransaction} from "@ledgerhq/hw-app-btc/lib/splitTransaction";
+import {serializeTransactionOutputs} from "@ledgerhq/hw-app-btc/lib/serializeTransaction";
+
 /**
  * Constant defining Ledger interactions.
  *
@@ -47,10 +50,6 @@ import {
  * @default ledger
  */
 export const LEDGER = 'ledger';
-import {splitTransaction} from "@ledgerhq/hw-app-btc/lib/splitTransaction";
-import {serializeTransactionOutputs} from "@ledgerhq/hw-app-btc/lib/serializeTransaction";
-
-const bitcoin = require('bitcoinjs-lib');
 
 const TransportU2F = require("@ledgerhq/hw-transport-u2f").default;
 const TransportWebUsb = require("@ledgerhq/hw-transport-webusb").default;
@@ -170,12 +169,26 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
       const transport = await TransportWebUsb.create();
       return callback(transport);
     } catch (e) {
-      if (e.message && e.message === "navigator.usb is undefined") { // webusb failed, so try u2f
-        const transport = await TransportU2F.create();
-        return callback(transport);
-      } else {
-        throw new Error("Error with Ledger transport");
+      if (e.message) {
+        //WebUSB Errors:
+        // If you hit cancel on the browser webusb popup
+        if (e.message === 'No device selected.') {
+          e.message = `Select your device in the WebUSB dialog box. Make sure it's plugged in, unlocked, and has the Bitcoin app open.`;
+        }
+        if (e.message === 'undefined is not an object (evaluating \'navigator.usb.getDevices\')') {
+          e.message = `Safari is not a supported browser.`;
+        }
+        // If you ask for a Transport in Firefox, fallback to U2F
+        if (e.message === "navigator.usb is undefined") {
+          try {
+            const transport = await TransportU2F.create();
+            return callback(transport);
+          } catch(err) {
+            throw new Error(err.message);
+          }
+        }
       }
+      throw new Error(e.message);
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/ledger.js
```diff
@@ -170,12 +169,26 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
...
-        throw new Error("Error with Ledger transport");
+      if (e.message) {
+        //WebUSB Errors:
+        // If you hit cancel on the browser webusb popup
+        if (e.message === 'No device selected.') {
+          e.message = `Select your device in the WebUSB dialog box. Make sure it's plugged in, unlocked, and has the Bitcoin app open.`;
+        }
+        if (e.message === 'undefined is not an object (evaluating \'navigator.usb.getDevices\')') {
+          e.message = `Safari is not a supported browser.`;
+        }
```
waldenraines, line 180: Should we be using `e.message`?  Seems like `statusCode` or `statusText` (badly named) would be less fragile:

Example:
```js
{
  name: "TransportStatusError",
  message: "Ledger device: Condition of use not satisfied (denied by the user?) (0x6985)",
  stack: "Error↵    at new Je (https://waldenraines.github.i…k.js?__WB_REVISION__=2f69d2e5e4a3a6900d27:2:6804)",
  statusCode: 27013,
  statusText: "CONDITIONS_OF_USE_NOT_SATISFIED"
}
```

bucko13: Huh, yeah it seems weird to reassign the property of the error thrown. At the very least, either duplicating to make a new error object, or just a new constant `message` that is thrown at the end. 

I think it would be really good if we had some custom Error types. This would definitely be out of scope for this PR, but something that I encountered as potentially useful when I was developing in caravan too.

humanumbrella: The errors you're listing are not being sanitized in any way from the calling application (caravan test suite) - so those are the actual messages that come back from the device interactions.

nikkirostollan: Agreed, statusCode generally is more reliable.  I think we had gotten in the habit of message checking b/c the didn't initially pass along error codes

humanumbrella: @nikkirostollan  -- The status codes and text will be visible once we're actually interfacing with the device. The error's we're attempting to catch here in calling the transport are because we cannot yet talk to the device at all.

E.g. error object looks like this, so message is what we have to work with:
Screen Shot 2020-07-15 at 10 12 35 AM

nikkirostollan: Ahh makes sense.

humanumbrella: For completeness: 
firefox error -> Screen Shot 2020-07-15 at 10 19 25 AM
cancel on webusb -> 
Screen Shot 2020-07-15 at 10 20 44 AM

src/ledger.js
```diff
@@ -170,12 +169,26 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
...
+        // If you ask for a Transport in Firefox, fallback to U2F
+        if (e.message === "navigator.usb is undefined") {
+          try {
+            const transport = await TransportU2F.create();
```
nikkirostollan, line 184: We have to make this switch in error handling because of the nano x, right? I'm asking to make sure it's not something we can decide by passing in additional params at instantiation.

src/ledger.js
waldenraines: Sorry for the confusion, I was just saying that checking `message` instead of `statusCode` (or whatever) seemed fragile because they could change the message at any point but the `statusCode` seemed less likely to change.

I'm not seeing the same errors you are in webusb.  

Device locked:
```js
{
  name: "TransportStatusError",
  message: "Ledger device: Internal error, please report (0x6f04)",
  stack: "Error↵    at new Je (https://waldenraines.github.i….js?__WB_REVISION__=0583eaf57b022e5b05f7:2:10903)",
  statusCode: 28420,
  statusText: "UNKNOWN_ERROR"
}
```

Clicking cancel on web usb dialog:
```js
Error: Error with Ledger transport
    at 2.170be0ce.chunk.js?__WB_REVISION__=0583eaf57b022e5b05f7:2
    at c (2.170be0ce.chunk.js?__WB_REVISION__=0583eaf57b022e5b05f7:2)
    at Generator._invoke (2.170be0ce.chunk.js?__WB_REVISION__=0583eaf57b022e5b05f7:2)
    at Generator.throw (2.170be0ce.chunk.js?__WB_REVISION__=0583eaf57b022e5b05f7:2)
    at p (2.170be0ce.chunk.js?__WB_REVISION__=0583eaf57b022e5b05f7:2)
    at s (2.170be0ce.chunk.js?__WB_REVISION__=0583eaf57b022e5b05f7:2)
```

Bitcoin app not open:

```js
{
  name: "TransportStatusError",
  message: "Ledger device: INS_NOT_SUPPORTED (0x6d00)",
  stack: "Error↵    at new Je (https://waldenraines.github.i….js?__WB_REVISION__=0583eaf57b022e5b05f7:2:10903)",
  statusCode: 27904,
  statusText: "INS_NOT_SUPPORTED"
}
```

Rejected sign:

```js
{
  name: "TransportStatusError",
  message: "Ledger device: Condition of use not satisfied (denied by the user?) (0x6985)",
  stack: "Error↵    at new Je (https://waldenraines.github.i….js?__WB_REVISION__=0583eaf57b022e5b05f7:2:10903)",
  statusCode: 27013,
  statusText: "CONDITIONS_OF_USE_NOT_SATISFIED"
}
```

Etc.


How did you cause the error in the screenshot?

src/ledger.js
humanumbrella: All of this was a hack to switch the Transport we call from webusb over to the U2F transport in the case of a Ledger device interaction request from Firefox. The Nano X does not show up in the device list of webusb via Chrome on Windows for ... 'complicated firmware reasons'. Therefore, Nano X on Windows would be unable to do anything without this fallback.

It's just a bonus that Nano S/X work in FF on other OSes with this change.

waldenraines: Yeah, because nano x doesn't work in Chrome on Windows 10 version 1909 but works in Windows 10 version 2004 (at least on my VM).

What Nikki is suggesting is cleaner but I'm not sure how difficult it would be.  If it's difficult then I'd personally be okay with this approach as u2f's days are numbered anyway.

src/ledger.js
humanumbrella: you need to run unchained-wallets locally, and then console.log the error to see it, e.g.
```
async withTransport(callback) {
    try {
      const transport = await TransportWebUsb.create();
      return callback(transport);
    } catch (e) {
      console.log(e);
      throw new Error(e);
    }
  }
```

There are two forms of webusb failures:  ones where you're already paired to the device, and ones where you're not yet paired to a device.  Also note: webusb can actually look like three separate *devices* to the browser: 

1. Pair with unlocked but on dashboard
2. Pair with Bitcoin app open
3. Pair with Bitcoin Testnet app open

The errors I am showing are when you're not yet paired with anything or when you're paired with unlock+dashboard and trying to do things ( ). The errors you're seeing is when you're paired with a bitcoin/testnet app and you're seeing actual errors.

To remove pairings, click on "Not Secure" in Chrome and click the x's next to the device pairings.
Screen Shot 2020-07-15 at 11 08 57 AM

As for `Error with Ledger transport` - that comes from https://github.com/unchained-capital/unchained-wallets/blob/v0.1.3/src/ledger.js#L177

The point is that the pre-paired errors (and only those) are what we're handling here. The post-paired errors are handled by the application calling it and account for almost all of the errors you are posting.

bucko13: Basically, anything that's happening in `withTransport` right since that's what's setting up the pairing?

humanumbrella: Yes - the other errors are firing from successfully using the `transport` but unsuccessfully talking to the particular `app`.

PR #31, opened 2020-07-21
Remove conditional logic to add SIGHASH_ALL byte

humanumbrella:
If the signature ends in `0x01` and the downstream application is expecting a SIGHASH_ALL byte to be included, then this conditional will invalidate the signature.

Make this API always return the signature with the SIGHASH byte included.

### DIFF
--- a/src/ledger.test.js
+++ b/src/ledger.test.js
@@ -218,6 +218,20 @@ describe('ledger', () => {
           });
         }
 
+        it("checks signatures include proper SIGHASH byte", () => {
+          // Signature format:
+          //   first byte signifies DER encoding           (0x30)
+          //   second byte is length of signature in bytes (0x03)
+          // The string length is however long the signature is minus these two starting bytes
+          // plain signature without SIGHASH (foobar is 3 bytes, string length = 6, which is 3 bytes)
+          expect(interactionBuilder().parse(["3003foobar"])).toEqual(["3003foobar01"]);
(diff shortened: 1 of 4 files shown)


bucko13:
I don’t feel comfortable approving this until [we know why this check was here in the first place](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence). If we can’t easily narrow this down but we test with a few different random txs and keys then I say we move forward since this is an important patch, but still think it warrants some investigation.

src/ledger.test.js
```diff
@@ -218,6 +218,20 @@ describe('ledger', () => {
           });
         }
 
+        it("checks signatures include proper SIGHASH byte", () => {
```
waldenraines, line 221: :+1:

src/trezor.js
```diff
@@ -683,7 +684,10 @@ export class TrezorSignMultisigTransaction extends TrezorInteraction {
    * @returns {string[]} array of input signatures, one per input
    */
   parse(payload) {
-    return (payload.signatures || []).map((inputSignature) => (`${inputSignature}01`));
+    // While we don't anticipate Trezor making firmware changes to include SIGHASH bytes with signatures,
+    // let's go ahead and make sure that we're not double adding the SIGHASH byte in case they do in the future.
```
waldenraines, line 688: :+1:

bucko13:
Code looks good to me. Good call adding for Trezor too.

PR #35, opened 2020-08-07
Improve ConfirmMultisigAddress with optional publicKey parameter

destrys:
# Summary
When calling `ConfirmMultisigAddress` the user should expect this behavior:
For Device A which is in control of Public Key A at BIP32Path A:
1. The multisig address should be shown on the Device A.
2. BIP32Path A of the Public Key A is shown on Device A.
3. If Public Key A isn't in the redeem script of the multisig address, an error is thrown.
4. If the key on Device A at BIP32Path A isn't Public Key A, an error is thrown.

Before this PR, only 1-3 was true. This PR adds 4 via an optional parameter, for backwards compatibility.

## Details
`ConfirmMultisigAddress` currently only works on Trezor.  
The trezor interaction uses [getAddress](https://github.com/trezor/connect/blob/v8/docs/methods/getAddress.md#exporting-single-address). The important parameters passed to `getAddress` are: `path`, `address` and `multisig`, which allows you to verify that the key at on the device at `path` is used in `address`.

The problem arises when you use two public keys from two devices at the same BIP32 path. In this scenario, the inputs to `getAddress` are the same, so the devices are interchangeable. This means you can't specify which specific device you are checking, just whether or not the device is in control of **a key**, not a **specific key**.

A better solution is for Trezor to have either a public key, key index (which key in the multisig redeemscript), or a device fingerprint parameter.

This PR works around that by getting *two* address in the `getAddress` call via the `bundle` optional parameter. If an optional `publicKey` parameter is passed, the *singlesig* address at the same bip32 path is retrieved (without showing on device). The singlesig address is compared to the `publicKey`, if they don't agree, an error is thrown.

A benefit of this approach is the the Trezor T will display the bip32 path (as long as it isn't a standard *singlesig* path), which it wouldn't otherwise. A downside to this approach is that the user confirms the address on the device but an error is still returned, which could cause some confusion.

### DIFF
--- a/src/trezor.js
+++ b/src/trezor.js
@@ -37,7 +37,9 @@ import {
   P2SH_P2WSH,
   P2WSH,
   signatureNoSighashType,
+  networkData,
 } from "unchained-bitcoin";
+import bitcoin from "bitcoinjs-lib";
 
 import {
   DirectKeystoreInteraction,
@@ -793,11 +795,13 @@ export class TrezorConfirmMultisigAddress extends TrezorInteraction {
    * @param {string} options.network - bitcoin network
    * @param {string} options.bip32Path - BIP32 path to the public key on this device used in the multisig address
    * @param {Multisig} options.multisig - multisig object
+   * @param {string} options.publicKey - optional public key to confirm
    */
-  constructor({network, bip32Path, multisig}) {
+  constructor({network, bip32Path, multisig, publicKey}) {
     super({network});
     this.bip32Path = bip32Path;
     this.multisig = multisig;
+    this.publicKey = publicKey;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


destrys:
Agreed, annoying the test suite is in caravan.
It might make sense to move the integration tests into this library and just have caravan skin them for end-user debugging.

src/trezor.js
```diff
@@ -864,23 +877,67 @@ export class TrezorConfirmMultisigAddress extends TrezorInteraction {
...
+      ]
+    }
   }
 
+  parse(payload) {
+    if (!this.publicKey) {
+      return payload;
+    }
+    const keyPair = bitcoin.ECPair.fromPublicKey(
+      Buffer.from(this.publicKey, 'hex'))
```
humanumbrella, line 930: ;

src/trezor.js
```diff
@@ -864,23 +877,67 @@ export class TrezorConfirmMultisigAddress extends TrezorInteraction {
...
+    const { address } = bitcoin.payments.p2pkh({
+      pubkey: keyPair.publicKey,
+      network: networkData(this.network),
+    })
```
humanumbrella, line 934: ;

src/trezor.js
```diff
@@ -864,23 +877,67 @@ export class TrezorConfirmMultisigAddress extends TrezorInteraction {
...
+          multisig: {
+            m: multisigRequiredSigners(this.multisig),
+            pubkeys: multisigPublicKeys(this.multisig).map((publicKey) => trezorPublicKey(publicKey)),
+          },
+          scriptType: ADDRESS_SCRIPT_TYPES[multisigAddressType(this.multisig)],
         },
-        scriptType: ADDRESS_SCRIPT_TYPES[multisigAddressType(this.multisig)],
-      },
-    ];
+      ]
```
humanumbrella, line 921: ;

PR #36, opened 2020-08-11
adds translator for creating coldcard multisig wallet config files fr…

humanumbrella:
…om caravan configs

Work in progress for Coldcard integration

### DIFF
--- a/src/coldcard.js
+++ b/src/coldcard.js
@@ -1,0 +1,103 @@
+/**
+ * Provides classes for interacting with a Coldcard via JSON/PSBT files
+ *
+ * @module coldcard
+ */
+import {
+  IndirectKeystoreInteraction,
+} from "./interaction";
+
+export const COLDCARD = 'coldcard';
+
+export function parseData() {}
+
+export function encodeData() {}
+
+/**
+ * Base class for interactions with Coldcard.
+ *
+ * @extends {module:interaction.IndirectKeystoreInteraction}
+ */
+export class ColdcardInteraction extends IndirectKeystoreInteraction {
+
+
+  /**
+   * Coldcard interactions need to know the bitcoin network they are on
+   *
+   * @param {object} options - options argument
+   * @param {string} options.network - bitcoin network
+   */
+  constructor({network}) {
+    super();
+    this.network = network;
+  }
+}
+
+/**
+ * Base class for interactions
+ *
+ * @extends {module:hermit.HermitInteraction}
+ */
+export class ColdcardFileReader extends ColdcardInteraction {}
+
+export class ColdcardExportExtendedPublicKey extends ColdcardFileReader {}
+
+export class ColdcardSignTransaction extends ColdcardFileReader {}
+
+/**
+ * Returns a valid multisig wallet config text file to send over to a Coldcard
+ *
+ * NOTE: while it will technically work with unknown root fingerprints (xfp's)
+ * we highly recommend only setting up the multisig wallet with complete xfp
+ * information. We define isSupported() to check exactly this condition.
+ *
+ * This is an example Coldcard config file from
+ * https://coldcardwallet.com/docs/multisig
+ *
+ * # Coldcard Multisig setup file (exported from 4369050F)
+ * #
+ * Name: MeMyself
+ * Policy: 2 of 4
+ * Derivation: m/45'
+ * Format: P2WSH
+ *
+ * D0CFA66B: tpubD9429UXFGCTKJ9NdiNK4rC5...DdP9
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardConfig {
+  constructor({jsonConfig}) {
+    this.jsonConfig = jsonConfig;
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


src/coldcard.js
```diff
@@ -0,0 +1,103 @@
+/**
+ * Provides classes for interacting with a Coldcard via JSON/PSBT files
```
destrys, line 2: The wallet config is technically neither of these formats =)

src/coldcard.js
```diff
@@ -0,0 +1,103 @@
...
+   */
+  translate() {
+    // Coldcard configs can't have spaces in the names, it just splits on space and takes the first word.
+    let output =
+`# Coldcard Multisig setup file (exported from unchained-wallets)
+# https://github.com/unchained-capital/unchained-wallets
+# 
+Name: ${this.jsonConfig.name.replace(/ /g, '_')}
+Policy: ${this.jsonConfig.quorum.requiredSigners} of ${this.jsonConfig.quorum.totalSigners}
+Derivation: m/45'
```
destrys, line 94: I don't think `m/45'` should be hardcoded in here. I'm not sure what the right value is.
Does it need to be the path for the key that is held by the ColdCard? 
If there are multiple xpubs from the same coldcard (same XFP), what happens?
Is it possible to just not include it at all? (I think this would be best if possible.)

src/coldcard.js
```diff
@@ -0,0 +1,103 @@
...
+ * Derivation: m/45'
+ * Format: P2WSH
+ *
+ * D0CFA66B: tpubD9429UXFGCTKJ9NdiNK4rC5...DdP9
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardConfig {
```
destrys, line 70: Is this getting used anywhere else? Might make more sense to have it just be a function instead of a class.
If it doesn't have everything it needs, it can just throw an error.

src/coldcard.js
```diff
@@ -0,0 +1,103 @@
...
+  isSupported() {
+    // Do we have all of the fingerprints?
+    return this.jsonConfig.extendedPublicKeys.every((xpub) => xpub.xfp !== 'Unknown');
+  }
+
+  /**
+   * @returns {string} configText - output to be written to a textfile and uploaded to Coldcard.
+   */
+  translate() {
+    // Coldcard configs can't have spaces in the names, it just splits on space and takes the first word.
```
destrys, line 87: probabaly should call `isSupported` here, otherwise there isn't a check that an XFP isn't 'Unknown'

src/index.js
```diff
@@ -340,7 +355,28 @@ export function ConfirmMultisigAddress({keystore, network, bip32Path, multisig,
...
 
+/**
+ * Return an interaction class for confirming a multisig addreess with
+ * the given `keystore`.
+ *
+ * @param {KEYSTORE} options.keystore - keystore to use
+ * @param {string} jsonConfig - asdf
+ * @returns {ColdcardConfig|UnsupportedInteraction} - text in shape of coldcard multisig wallet import
+ */
+export function translateConfig({keystore, jsonConfig}) {
```
destrys, line 366: If we want to put config -> electrum or other coordinators, this won't make sense.
But then again, if there are a class of keystores that have config files, this will make sense.
I guess this works.

src/coldcard.js
humanumbrella: ah right, i'm checking isSupported in the consuming application - should do it here instead.

src/coldcard.js
humanumbrella: TXT

src/coldcard.js
humanumbrella: still some room for discussion here - factored it out to constructor. this is enough to get us started.

can submit without derivation ... could still sign. not sure about multiple xpubs from the same XFP yet

src/coldcard.js
```diff
@@ -0,0 +1,108 @@
...
+ * D0CFA66B: tpubD9429UXFGCTKJ9NdiNK4rC5...DdP9
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardConfig {
+  constructor({jsonConfig}) {
+    this.jsonConfig = jsonConfig;
+    //this.jsonConfig.derivation = "m/45'";
```
bucko13, line 73: Does this comment need to be cleaned up or converted to a TODO?

src/coldcard.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * @returns {boolean} isSupported - does the jsonConfig have necessary pieces?
+   */
+  isSupported() {
```
bucko13, line 79: To me this feels like a check (and a rejection) that should happen at the instantiation of a new class object. If when creating a new instance of a `ColdcardConfig` the required information is missing (maybe we're not requiring it and just want to log a warning? not sure how we handle "suggestions" but we're throwing in `translate` if this is missing so we're treating it as required so far) then the class should never be instantiated in the first place. So this to me should be in the constructor imo.

humanumbrella:
More discussions unfolding - **do not merge**.

PR #37, opened 2020-08-17
XFP should be 8-char hex

destrys:
The XFP is converted from an int to a hex string, but wasn't being zero-padded to 8 characters, so there was a 1-in-16 chance of a 7 character XFP, and a 1-in-256 chance of a 6-digit XFP.

This PR adds the necessary zero-padding.

### DIFF
--- a/src/trezor.js
+++ b/src/trezor.js
@@ -488,7 +488,9 @@ export class TrezorExportHDNode extends TrezorInteraction {
     for (let i = 0; i < payload.length; i++) {
       // Find the payload with bip32 = MULTISIG_ROOT to get xfp
       if (payload[i].serializedPath === MULTISIG_ROOT) {
-        rootFingerprint = payload[i].fingerprint.toString(16);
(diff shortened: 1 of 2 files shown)


src/trezor.js
```diff
@@ -488,7 +488,9 @@ export class TrezorExportHDNode extends TrezorInteraction {
     for (let i = 0; i < payload.length; i++) {
       // Find the payload with bip32 = MULTISIG_ROOT to get xfp
       if (payload[i].serializedPath === MULTISIG_ROOT) {
-        rootFingerprint = payload[i].fingerprint.toString(16);
+        let fp = payload[i].fingerprint;
+        // zero pad the hex string
+        rootFingerprint = (fp + 0x100000000).toString(16).substr(-8);
```
waldenraines, line 493: > I will be adding additional tests, but I have used a previously failing input and this solves the problem.

@humanumbrella as part of adding tests, can we factor this out?

PR #40, opened 2020-09-04
Adds configuration file adapter

humanumbrella:
This will allow the transforming of Unchained and Caravan multisig config files into Coldcard multisig config files. Some unit tests to make sure config files have necessary pieces are included.

### DIFF
--- a/src/coldcard.js
+++ b/src/coldcard.js
@@ -1,0 +1,150 @@
+/**
+ * Provides classes for interacting with a Coldcard via TXT/JSON/PSBT files
+ *
+ * The following API classes are implemented:
+ *
+ * * ColdcardExportExtendedPublicKey
+ * * ColdcardSignMultisigTransaction
+ * * ColdcardMultisigWalletConfig
+ *
+ * @module coldcard
+ */
+import {
+  IndirectKeystoreInteraction,
+} from "./interaction";
+
+export const COLDCARD = 'coldcard';
+export const WALLET_CONFIG_VERSION = "0.0.1";
+
+
+/**
+ * Base class for indirect interactions with Coldcard.
+ *
+ * @extends {module:interaction.IndirectKeystoreInteraction}
+ */
+export class ColdcardInteraction extends IndirectKeystoreInteraction {
+
+}
+
+/**
+ * Base class for coldcard interactions (reading/writing files)
+ *
+ * @extends {module:hermit.HermitInteraction}
+ */
+export class ColdcardFileReader extends ColdcardInteraction {
+
+}
+
+export class ColdcardExportExtendedPublicKey extends ColdcardFileReader {
+
+}
+
+/**
+ * Returns signature request data to display via PSBT for Coldcard and returns the PSBT
+ * data back to the calling application
+ */
+export class ColdcardSignTransaction extends ColdcardFileReader {
+
+}
+
+/**
+ * Returns a valid multisig wallet config text file to send over to a Coldcard
+ *
+ * NOTE: only the root xfp of the signing device is required to be correct, but we
+ * recommend only setting up the multisig wallet on the Coldcard with complete xfp
+ * information. We define isSupported() to check exactly this condition.
+ *
+ * This is an example Coldcard config file from
+ * https://coldcardwallet.com/docs/multisig
+ *
+ * # Coldcard Multisig setup file (exported from 4369050F)
+ * #
+ * Name: MeMyself
+ * Policy: 2 of 4
+ * Derivation: m/45'
+ * Format: P2WSH
+ *
+ * D0CFA66B: tpubD9429UXFGCTKJ9NdiNK4rC5...DdP9
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardMultisigWalletConfig {
+  constructor({jsonConfig}) {
+    if (typeof jsonConfig === "object") {
+      this.jsonConfig = jsonConfig;
+    }else if (typeof jsonConfig === "string") {
+      try {
+        this.jsonConfig = JSON.parse(jsonConfig);
+      } catch (error) {
+        throw new Error("Unable to parse JSON.");
+      }
+    } else {
+      throw new Error("Not valid JSON.")
+    }
+
+    if ((this.jsonConfig.uuid || this.jsonConfig.name)) {
+      this.name = this.jsonConfig.uuid || this.jsonConfig.name;
+    } else {
+      throw new Error("Configuration file needs a UUID or a name.");
+    }
+
+    if (this.jsonConfig.quorum.requiredSigners && this.jsonConfig.quorum.totalSigners) {
+      this.requiredSigners = this.jsonConfig.quorum.requiredSigners;
+      this.totalSigners = this.jsonConfig.quorum.totalSigners;
+    } else {
+      throw new Error("Configuration file needs quorum.requiredSigners and quorum.totalSigners.");
+    }
+
+    if (this.jsonConfig.addressType) {
+      this.addressType = jsonConfig.addressType;
+    } else {
+      throw new Error("Configuration file needs addressType.")
+    }
+
+    if (this.jsonConfig.extendedPublicKeys &&
+      this.jsonConfig.extendedPublicKeys.every((xpub) => {
+        // For each xpub, check that xfp exists, the length is 8, type is string, and valid hex
+        if (!xpub.xfp) {
+          throw new Error("ExtendedPublicKeys missing at least one xfp.");
+        }
+        if (typeof xpub.xfp !== 'string') {
+          throw new Error("XFP not a string");
+        }
+        if (xpub.xfp.length !== 8) {
+          throw new Error("XFP not length 8");
+        }
+        if (isNaN(Number(`0x${xpub.xfp}`))) {
+          throw new Error("XFP is invalid hex");
+        }
+        return true;
+      })) {
+      this.extendedPublicKeys = this.jsonConfig.extendedPublicKeys;
+    } else {
+      throw new Error("Configuration file needs extendedPublicKeys.")
+    }
+  }
+
+  /**
+   * @returns {string} output to be written to a text file and uploaded to Coldcard.
+   */
+  adapt() {
+    // Coldcard configs can't have spaces in the names, it just splits on space and takes the first word.
+    // Currently operating without derivation paths per xpub until feature is added.
+    let output = `# Coldcard Multisig setup file (exported from unchained-wallets)
+# https://github.com/unchained-capital/unchained-wallets
+# v${WALLET_CONFIG_VERSION}
+# 
+Name: ${this.name}
+Policy: ${this.requiredSigners} of ${this.totalSigners}
+Format: ${this.addressType}
+
+`;
+    // We need to loop over xpubs and output `xfp: xpub` for each
+    let xpubs = this.extendedPublicKeys.map((xpub) => `${xpub.xfp}: ${xpub.xpub}`);
+    output += xpubs.join("\r\n");
+    output += "\r\n";
+    return output;
+  }
+}
--- a/src/coldcard.js
+++ b/src/coldcard.js
@@ -74,7 +74,7 @@ export class ColdcardMultisigWalletConfig {
   constructor({jsonConfig}) {
     if (typeof jsonConfig === "object") {
       this.jsonConfig = jsonConfig;
-    }else if (typeof jsonConfig === "string") {
+    } else if (typeof jsonConfig === "string") {
       try {
         this.jsonConfig = JSON.parse(jsonConfig);
       } catch (error) {
--- a/src/coldcard.test.js
+++ b/src/coldcard.test.js
@@ -1,0 +1,149 @@
+import {ColdcardMultisigWalletConfig, WALLET_CONFIG_VERSION} from './coldcard';
+
+describe("ColdcardMultisigWalletConfig", () => {
+  const jsonConfigUUID = {
+    name: "Test",
+    addressType: "P2SH",
+    network: "testnet",
+    quorum: {
+      requiredSigners: 2,
+      totalSigners: 3
+    },
+    startingAddressIndex: 5,
+    extendedPublicKeys: [
+      {
+        name: "unchained",
+        xpub: "tpubDF17mBZYUi35r7UEkGr7SjSisec8QL2J1zGh9WQtmnhJtHeMFy2eH2Xsnr2ynxENbqHmcEA4NnoT8T6RZxks4G5evZdWy1RbSPTm8LtNPU3",
+        bip32Path: "Unknown",
+        xfp: "83471e78"
+      },
+      {
+        name: "Os_words1",
+        xpub: "tpubDF61GHbPYRhEEsqNTDF2jbMBkEMmoksy1h2URbhZ8p7JfR9QRgbn6vkeA7g3t4Ue6uSHhYJxD9mRVz1ZQVYyW3RAPPuwVM4UeZyZPKu89DY",
+        bip32Path: "m/45'/1'/50'/0",
+        xfp: "f57ec65d"
+      },
+      {
+        name: "Os_words2",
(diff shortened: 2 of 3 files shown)


src/coldcard.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * @returns {string} output to be written to a text file and uploaded to Coldcard.
+   */
+  adapt() {
+    // Coldcard configs can't have spaces in the names, it just splits on space and takes the first word.
+    // Currently operating without derivation paths per xpub until feature is added.
+
+    if ((this.jsonConfig.uuid || this.jsonConfig.name) &&
```
bucko13, line 91: Should these field validations be in the constructor instead? If certain fields are missing at initialization then surely it's an invalid config which should throw before anything with it is used. This also avoids having to do any validation with any other methods we might have and we can assume that an initialized `ColdcardMultisigConfig` class is good to use.

src/coldcard.js
```diff
@@ -0,0 +1,113 @@
...
+ * D0CFA66B: tpubD9429UXFGCTKJ9NdiNK4rC5...DdP9
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardMultisigWalletConfig {
+  constructor({jsonConfig}) {
+    if (jsonConfig &&
+      jsonConfig.extendedPublicKeys &&
```
bucko13, line 76: Why leave these in the `jsonConfig`? to me it feels like once initialized from the json that's passed in, these become properties of the config itself. `configInstance.jsonconfig.extendedPublicKeys` feels a little unwieldy if you're trying to access. This becomes even nicer for internal access, e.g. `this.extendedPublicKeys`. In terms of how it reads, it feels more natural too (i.e. "these are the xpubs of this coldcard config class")

src/coldcard.js
```diff
@@ -0,0 +1,113 @@
...
+ * Derivation: m/45'
+ * Format: P2WSH
+ *
+ * D0CFA66B: tpubD9429UXFGCTKJ9NdiNK4rC5...DdP9
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardMultisigWalletConfig {
```
bucko13, line 73: Something that feels like it would be nice to have are methods that are named like `fromColdcardConfig` and `toColdcardConfig`. The `from` would be a tool to parse what is retrieved from the config file and turn it into a class that has some actual functionality. The `to` method is basically what I'm gathering `adapt` does. 

But maybe this is more of an argument for what's been discussed elsewhere about having a general purpose `MultisigWalletConfig` class which in addition to the above methods could be a lot more useful and flexible and include things like `toElectrumConfig` (and `from`), `toCaravanConfig`, etc. Doing it this way has a nice added benefit of once you've imported/converted, uc-wallets/bitcoin is then providing a uniform interface for interacting with a multisig wallet config that can be passed around and converted as necessary. 

Starting with what's here as a base and just generalizing the naming, even if it only works for coldcard for now shouldn't involve too much while offering more flexibility for the long term.

src/coldcard.test.js
```diff
@@ -0,0 +1,175 @@
...
+      {
+        "name": "Os_words2",
+        "xpub": "tpubDEZxsxeamoPim6T1hg2igm4vDmdQrQnHVH5gM5NjcjowtgYVv5ZhhR5sFgRVNjRFGk1HmxsFsYhu3jGaAGVCpCsL5AbAVk6xKssr6gK3tPk",
+        "bip32Path": "m/45'/1'/60'/0",
+        "xfp": "f57ec65d",
+      },
+    ],
+    "uuid": "OWPyFOA1"
+  };
+  const jsonConfigName = {
```
bucko13, line 35: Any reason why these JSON objects need to be duplicated almost exactly? Since there's already some parsing/stringifying and copying below, couldn't there just be one as a base that is adjusted as necessary for each specific test?

src/coldcard.js
humanumbrella: This sounds good on the surface but I don't how it would work in practice given the missing information (bip32 paths), for example, in the current Coldcard config. The `fromColdcardConfig` e.g. does not give enough information for us to create a CaravanConfig or an ElectrumConfig.

bucko13: But then this can be handled in the conversion. If you import a config via `fromColdcardConfig`, we could then have methods to add missing information, which the consumer of this library could decide any number of ways they want to retrieve it. Then if a `.toElectrumConfig` is called if it's not available on the current object, you would get an error (and a very specific error telling you why an electrum config isn't possible to generate.

bucko13: I actually had to do something similar on another project when dealing with LSATs (lightning service authentication tokens). Basically there were a lot of different ways to consume and generate lsats. They are typically, like the Coldcard Config from what I can tell, sent in http headers as text encoded in a particular way. But I needed a way so that specific information could be consumed and easily read. So, this tool would read the plain text, learn where things were broken up and then add as properties on the object. If you wanted to convert back as an LSAT or as a challenge, the conversion method determines if it has enough information and throws a descriptive error if it doesn't. 

Here's the main file I'm thinking about: https://github.com/Tierion/lsat-js/blob/master/src/lsat.ts
You can see in [fromChallenge](https://github.com/Tierion/lsat-js/blob/master/src/lsat.ts#L347) it will look for all the text elements it expects.
Meanwhile [toChallenge](https://github.com/Tierion/lsat-js/blob/master/src/lsat.ts#L245) throws if it doesn't have an invoice property attached to it.

Anyway, sorry to bring a totally random and unrelated project into this. It's just something I was working on relatively recently and felt relevant. In short, I think the ability for this Class to actually have the ability to tell us it can't be converted to an ElectrumConfig is super powerful as a tool and feels like it belongs together with the other conversion classes since the config, unlike the keystore classes, are kind of supposed to be interoperable in my opinion.

humanumbrella: Yeah - I hear that. I'm open to more discussion 1on1 or with a wider audience, but I'd argue that if you want an Electrum config from a Coldcard, there's a path on the device for that (Multisig Wallet Details > Electrum Wallet) and it'll output the file you want to the SD card.

src/coldcard.js
```diff
@@ -0,0 +1,115 @@
...
+      jsonConfig.addressType &&
+      jsonConfig.extendedPublicKeys &&
+      jsonConfig.extendedPublicKeys.every((xpub) => xpub.xfp && xpub.xfp !== "Unknown")) {
+      this.extendedPublicKeys = jsonConfig.extendedPublicKeys;
+      this.name = jsonConfig.uuid || jsonConfig.name;
+      this.requiredSigners = jsonConfig.quorum.requiredSigners;
+      this.totalSigners = jsonConfig.quorum.totalSigners;
+      this.addressType = jsonConfig.addressType;
+    } else {
+      throw new Error("Configuration file must be JSON, " +
```
bucko13, line 88: Is this error technically correct? If it's JSON it would be text right? and then it would have to be parsed first in order to access the properties. In either case, it might be a good idea to have some type checks, support both, and then attempt to `JSON.parse` if it's a string, catching an error to say it's mis-configured.

src/coldcard.js
```diff
@@ -0,0 +1,115 @@
...
+ */
+export class ColdcardMultisigWalletConfig {
+  constructor({jsonConfig}) {
+    if (jsonConfig &&
+      (jsonConfig.uuid || jsonConfig.name) &&
+      jsonConfig.quorum.requiredSigners &&
+      jsonConfig.quorum.totalSigners &&
+      jsonConfig.addressType &&
+      jsonConfig.extendedPublicKeys &&
+      jsonConfig.extendedPublicKeys.every((xpub) => xpub.xfp && xpub.xfp !== "Unknown")) {
```
bucko13, line 81: How about checking if it's a hex rather than that it is not 'Unknown'. If there's a length restriction, that could be added as well.

src/coldcard.js
```diff
@@ -0,0 +1,115 @@
...
+ * D0CFA66B: tpubD9429UXFGCTKJ9NdiNK4rC5...DdP9
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardMultisigWalletConfig {
+  constructor({jsonConfig}) {
+    if (jsonConfig &&
+      (jsonConfig.uuid || jsonConfig.name) &&
```
bucko13, line 76: This is partly just a personal taste thing, but I think it's more readable to do these conditionals one at a type (unless they are dependent on each other).

i.e. 
```
if (jsonConfig.name && jsonConfig.uuid)
  this.name = jsonConfig.uuid || jsonConfig.name
```
An added benefit of this is that if there are any specific properties that are absolutely required vs some that aren't, you can throw in that conditional and say what's missing.

bucko13: Also for this section, do you have any feelings on also including type checking? Even though JS obviously isn't strictly typed, you can add asserts that will throw to make sure everything is kosher. This also makes this initialization a very effective validator pretty much out of the gate (which you want anyway since the class is useless if any field is not formatted correctly).

src/coldcard.test.js
```diff
@@ -0,0 +1,148 @@
+import {ColdcardMultisigWalletConfig, WALLET_CONFIG_VERSION} from './coldcard';
+
+describe("ColdcardMultisigWalletConfig", () => {
+  const jsonConfigUUID = {
```
bucko13, line 4: Two other suggestions:
1. I'm not sure the quotations around the key names are necessary since this is still technically an object and not a JSON string, so it could just make this easier to maintain.
2. It might be a good idea to put this in a `beforeEach` block so that the value is reset after each test to avoid tests polluting each other.

src/coldcard.js
bucko13: What about for someone not using a coldcard though? What if I want to go from Electrum to Caravan? It also could be useful in other UI settings.

bucko13: What if you have a UI like in caravan that lets you compose a config that you want to be able to import into either Coldcard or Electrum? You could have the form just pass the object to this class and depending on what is selected, spit out the correctly formatted text/file for Caravan, Coldcard, Electrum, or any other format that ends up getting created which we could extend this base class to support.

src/coldcard.test.js
humanumbrella: As for 1 - noted - I just pasted in what comes from unchained external spend JSON. updated
2 - definitely.

src/coldcard.test.js
```diff
@@ -0,0 +1,149 @@
...
+    addressType: "P2SH",
+    network: "testnet",
+    quorum: {
+      requiredSigners: 2,
+      totalSigners: 3
+    },
+    startingAddressIndex: 5,
+    extendedPublicKeys: [
+      {
+        name: "unchained",
```
bucko13, line 15: this might've been causing the issues with the spread operator since it's not a deep copy and these are objects within objects.

waldenraines:
@humanumbrella @bucko13 what is the status of this?

src/coldcard.test.js
```diff
@@ -0,0 +1,149 @@
...
+      {
+        name: "Os_words2",
+        xpub: "tpubDEZxsxeamoPim6T1hg2igm4vDmdQrQnHVH5gM5NjcjowtgYVv5ZhhR5sFgRVNjRFGk1HmxsFsYhu3jGaAGVCpCsL5AbAVk6xKssr6gK3tPk",
+        bip32Path: "m/45'/1'/60'/0",
+        xfp: "f57ec65d"
+      }
+    ],
+     uuid: "OWPyFOA1"
+  };
+  const coldcardConfigUUID = `# Coldcard Multisig setup file (exported from unchained-wallets)
```
bucko13, line 35: This doesn't need to be done in this PR necessarily, but I wonder if these would be useful in a fixtures file/directory? Might be something that's used later in other tests. Wouldn't even be surprised if there were some in the coldcard repo, which would be good to standardize on.

Not a blocking comment, just something that occurred to me looking at this again.

src/coldcard.test.js
```diff
@@ -0,0 +1,149 @@
...
+Name: Test
+Policy: 2 of 3
+Format: P2SH
+
+83471e78: tpubDF17mBZYUi35r7UEkGr7SjSisec8QL2J1zGh9WQtmnhJtHeMFy2eH2Xsnr2ynxENbqHmcEA4NnoT8T6RZxks4G5evZdWy1RbSPTm8LtNPU3\r
+f57ec65d: tpubDF61GHbPYRhEEsqNTDF2jbMBkEMmoksy1h2URbhZ8p7JfR9QRgbn6vkeA7g3t4Ue6uSHhYJxD9mRVz1ZQVYyW3RAPPuwVM4UeZyZPKu89DY\r
+f57ec65d: tpubDEZxsxeamoPim6T1hg2igm4vDmdQrQnHVH5gM5NjcjowtgYVv5ZhhR5sFgRVNjRFGk1HmxsFsYhu3jGaAGVCpCsL5AbAVk6xKssr6gK3tPk\r
+`;
+
+  let jsonConfigCopy = JSON.parse(JSON.stringify(jsonConfigUUID));
```
bucko13, line 61: This can just be initialized empty. `let jsonConfigCopy;` looks a little cleaner and is pretty standard in my experience. It'll get assigned in the beforeEach.

bucko13 (commented):
Just some small nits, really just optional comments. I still think it's worth discussing the Config class further, but otherwise this LGTM (and if we decide to keep it in the current format, then this also LGTM).

src/coldcard.test.js
humanumbrella: Yes I think updating the fixtures in `unchained-bitcoin` with PSBT and signed PSBT will be necessary and improve confidence with known transactions, potentially even add it to the test suite.

src/coldcard.test.js
humanumbrella: Got it

waldenraines (approved):
LGTM but will wait on @bucko13's approval before merging.

src/coldcard.test.js
bucko13: Would you mind adding this as an issue ticket to unchained-bitcoin and mentioning what specifically needs to be moved over? It might be best once this is merged in so there's something to link to but just wanted to mention here so we don't forget about it.

bucko13: I'd even support starting by adding a fixtures file here that we can eventually migrate to unchained-bitcoin when psbt support gets better built out there. Again, it shouldn't be a blocker necessarily, but I could see this being perpetually de-prioritized and more likely to have more constants that should be in fixtures added in as we continue building out.

src/coldcard.js
```diff
@@ -0,0 +1,150 @@
...
+ * 8E697B74: tpubD97nVL37v5tWyMf9ofh5rzn...XgSc
+ * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+ * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+ *
+ */
+export class ColdcardMultisigWalletConfig {
+  constructor({jsonConfig}) {
+    if (typeof jsonConfig === "object") {
+      this.jsonConfig = jsonConfig;
+    }else if (typeof jsonConfig === "string") {
```
bucko13, line 77: Marked this in the other PR (sorry!), but nit around the `else`

bucko13:
I think there are some overlapping comments now from the other PR #41 (mainly around fixtures, some formatting, and jsdoc comments) but in addition I do think the comments around the configuration adapter still warrants a discussion. I feel pretty strongly about this being more general functionality and think it would provide more long term utility creating such a tool. If there's a deeper discussion around this and we still come out on the side of keeping this a ColdCard specific tool, then let's move forward that way, but I still think it warrants a discussion where we're making a fully informed decision one way or another because this will more likely than not be the decision we're stuck with in the future (i.e. if we change our mind in the future, it's unlikely we'd spend the time unwinding). In the hopes of making this a utility that has appeal and utility for broad (but still targeted) use cases, it feels worth it to make sure we're thinking about how we want this to be used before committing to one approach or another.

bucko13 (approved):
@waldenraines there are some other comments that are getting addressed in #41 so this one should be good to go.

PR #41, opened 2020-09-25
Adds interaction classes and unit tests for Coldcard

humanumbrella:
Builds on work from https://github.com/unchained-capital/unchained-wallets/pull/40 so let's get that merged first.

Adds ability to parse JSON files with XPUBs from the Coldcard and ability to extract signatures from signed PSBTs using new functionality available in `unchained-bitcoin`


humanumbrella:
Looks like I need a release from `unchained-bitcoin` and then a bump to package version here before it will past tests.

package-lock.json
```diff
@@ -3409,28 +3414,28 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": false,
+          "resolved": "",
```
waldenraines, line 3417: Wonder what's up with all of these

src/coldcard.js
```diff
@@ -0,0 +1,431 @@
...
+    let xfpFromWithinXpub;
+
+    if (!bip32Path) {
+      throw new Error("No BIP32 path in JSON file.");
+    }
+    if (!rootFingerprint && xpubClass.depth !== 1) {
+      throw new Error("No xfp in JSON file.");
+    }
+
+    /* istanbul ignore else */
```
waldenraines, line 118: But shouldn't we handle the else here?

waldenraines (commented):
Just a couple of questions

src/coldcard.js
humanumbrella: No. This is for the upcoming else on `xpubClass.depth === 1` (L121) ... which, if the depth is not 1, we don't have any way to compute the fingerprint, so we should not do anything.

src/coldcard.test.js
```diff
@@ -0,0 +1,609 @@
...
+            '3045022100a41cd24ac399c24449334d0a46906626de003e2b88a9c37e46f2cb0cbfbac0f602203c7524ec1c8f12d535a2212632865da1bef43bbe81c1b20ead12b6cbf7f44f9a01',
+          ],
+      };
+      expect(result).toEqual(expectedOutput);
+      expect(Object.keys(result).length).toEqual(1);
+    });
+    it("return multi input, double signature set", () => {
+      const interaction = interactionBuilder({});
+      const result = interaction.parse(multiInputB64PSBT_fullySigned);
+      const expectedOutput = {
```
bucko13, line 355: I also feel like these would make sense in the fixtures file next to the strings that they are expected to be extracted from, but I could be misunderstanding these pieces. It would make the most sense if the outputs are the same regardless of what's doing the interaction, which it feels like should be the case here, but I could be wrong about that.

src/index.js
```diff
@@ -159,19 +183,19 @@ export function ExportExtendedPublicKey({
...
+      return new TrezorExportExtendedPublicKey({bip32Path,
         network,
-        bip32Path,
-        includeXFP,
-      });
+        includeXFP});
     case HERMIT:
       return new HermitExportExtendedPublicKey({bip32Path});
+    case COLDCARD:
+      return new ColdcardExportExtendedPublicKey({bip32Path,
```
bucko13, line 192: I'm surprised the linter isn't picking these up but the props being stacked should be formatted as they are further down in line 288

```
      return new ColdcardSignMultisigTransaction({
        network,
        inputs,
        outputs,
        bip32Paths,
        psbt,
      });
```

bucko13 (changes requested):
I got a little confused about the two PRs and didn't realize that this is building on top of the other PR, so, apologies if there are overlapping or redundant comments. 

There are some nits as well just general questions in here. One thing that I think should be added to stay consistent with the other classes is more jsdoc documentation on the classes and the individual methods. I think there are some in particular that would be useful given the aspects that are unique to ColdCard. 

I also think the tests could really benefit from constants getting split out into fixtures. This would not only help declutter the test files but also help maintainability (or even portability if we move them out) as well as improve the readability of the assertions (it's easier to read the test as what you expect something to equal, e.g. `fixtures.keyPath.xpub`, rather a string of characters)

src/coldcard.js
humanumbrella: Yes - it's just for test coverage.
This whole section is really a sanity check that mirrors what Coldcard does with their wallet config file. If you send in a config file with just an xpub, it will try to calculate the xfp if the xpub is depth=1. 
It's not strictly necessary to do this test, more of a might as well check it since we happen to have all the pieces. There's nothing to check if it's not there and there's no reason to *test* that we didn't check nothing.

package-lock.json
humanumbrella: I'm not entirely sure.

src/coldcard.js
```diff
@@ -0,0 +1,431 @@
...
+
+    // Sanity check if you send in a depth one xpub, we should get the same fingerprint
+    if ((xfpFromWithinXpub && rootFingerprint) &&
+      xfpFromWithinXpub !== rootFingerprint) {
+      throw new Error("Computed fingerprint does not match the one in the file.");
+    }
+
+    return {
+      xpub,
+      rootFingerprint: rootFingerprint || xfpFromWithinXpub
```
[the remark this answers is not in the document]
humanumbrella: Yes - this method takes what was sent in via the JSON file and returns an object with both the `xpub` (renamed from JSON because it's listed as a different property) and adds on `rootFingerprint` which can either be included in the JSON (default) or computed from the xpub if it's depth==1 (bonus / sanity check)

This is specific to the format of the coldcard file so I'm unsure why it would go in `unchained-bitcoin` rather than here. The computation of xfp from depth1 xpub is using functionality from `unchained-bitcoin`.

src/coldcard.js
```diff
@@ -0,0 +1,431 @@
...
+      throw new Error("Not valid JSON.");
+    }
+    if (Object.keys(this.xpubJSONFile).length === 0) {
+      throw new Error("Empty JSON file.");
+    }
+    return this.handleKeyExtraction(this.xpubJSONFile);
+    }
+
+  handleKeyExtraction(incomingJSON) {
+    let xpub = incomingJSON.p2sh;
```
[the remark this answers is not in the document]
humanumbrella: you're right here. Fixed

src/coldcard.js
```diff
@@ -0,0 +1,431 @@
...
+
+  /**
+   * Request for the PSBT data that needs to be signed.
+   * @returns {Object} Returns the local unsigned PSBT from transaction details
+   */
+  request() {
+    if (this.psbt) {
+      return this.psbt;
+    } else {
+      // TODO:  use unchained-bitcoin to build the PSBT from the other parameters we need to return
```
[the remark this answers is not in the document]
humanumbrella: This one can't be lost track of because implementing Caravan functionality will require addressing it, hehe.

src/coldcard.js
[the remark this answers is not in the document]
bucko13: Yeah, the computation is what I meant. I guess what I'm asking then is why do we need a special method to rename a property and return a computed value that is computed elsewhere. If we need an object with these property names, a whole method seems superfluous imo. If the main point of this method is to sanity check all the values that the coldcard config needs, then couldn't it focus on just doing that and be named something like `validateConfigs`? if we still want a method to format that object, then separating out the validation from the object formatting feels appropriate.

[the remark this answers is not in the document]
humanumbrella: I think this may have been a holdover from when I was expecting a variety of uses from the `FileReader` instead of simply a single JSON file. I have edited for clarity.

src/coldcard.test.js
```diff
@@ -1,95 +1,433 @@
...
+  ColdcardMultisigWalletConfig,
+  COLDCARD_BASE_BIP32,
+} from './coldcard';
+import {MAINNET, TESTNET} from "unchained-bitcoin";
+import {
+  INFO,
+  PENDING,
+  ACTIVE,
+} from './interaction';
+import {fixtures} from './fixtures';
```
bucko13, line 15: 😍

bucko13 (approved):
🎉

PR #43, opened 2020-10-05
Slight Coldcard Update

humanumbrella:
Refactoring a little bit to make things clearer.

- Rename FileParser because at this time it's only handling the xpub json (multisig settings)
- Said parser should be much less opinionated about what it returns
- Move opinion-forming functionality into parents


src/coldcard.js
```diff
@@ -211,26 +220,17 @@ export class ColdcardExportPublicKey extends ColdcardFileParser {
...
-    });
-    return messages;
+    return super.messages();
   }
 
   parse(xpubJSONFile) {
     const result = super.parse(xpubJSONFile);
-    let xpub = super.deriveXpubIfNecessary(result.xpub);
-
+    let xpub = super.deriveXpubIfNecessary(result.p2sh);
```
bucko13, line 228: Is the `super` necessary here? This class is already inheriting from the parent and so it should already have `this.deriveXpubIfNecessary` unless the parent's version of it is different then what's on this child class? (super is necessary for `parse` since we're in this class' own custom `parse`)

src/coldcard.js
```diff
@@ -211,26 +220,17 @@ export class ColdcardExportPublicKey extends ColdcardFileParser {
...
-      level: INFO,
-      code: "coldcard.export",
-      text: "Go to Settings > Multisig Wallets > Export XPUB",
-    });
-    return messages;
+    return super.messages();
   }
 
   parse(xpubJSONFile) {
     const result = super.parse(xpubJSONFile);
```
bucko13, line 227: You could simplify this method even more by doing `const { p2sh, rootFingerprint } = super.parse(xpubJSONFile)`. No big deal, but just thought about it since then line 233 doesn't need to change.

src/coldcard.js
```diff
@@ -276,25 +276,16 @@ export class ColdcardExportExtendedPublicKey extends ColdcardFileParser {
...
```
bucko13, line 283: same destructuring comment here if you wanted to implement that.

src/coldcard.js
```diff
@@ -325,60 +316,63 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
+   * @param {object} options.psbt - PSBT of the transaction to sign
    */
   constructor({network, inputs, outputs, bip32Paths, psbt}) {
     super();
-    this.network = network;
-    this.inputs = inputs;
-    this.outputs = outputs;
-    this.bip32Paths = bip32Paths;
-    this.psbt = psbt;
+    if (psbt) {
```
bucko13, line 323: what about just 
```
if (!psbt) {
 throw new Error('...');
}
...
```
makes it more readable imo, especially if this constructor grows (e.g. to do some validation)

src/coldcard.js
```diff
@@ -409,9 +410,13 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
 /**
  * Returns a valid multisig wallet config text file to send over to a Coldcard
  *
- * NOTE: only the root xfp of the signing device is required to be correct, but we
- * recommend only setting up the multisig wallet on the Coldcard with complete xfp
- * information. We define isSupported() to check exactly this condition.
+ * NOTE: technically only the root xfp of the signing device is required to be
+ * correct, but we recommend only setting up the multisig wallet on the Coldcard
+ * with complete xfp information. Here we actuall turn this recommendation into a
```
bucko13, line 415: typo: `actually`

src/coldcard.test.js
```diff
@@ -175,14 +134,25 @@ describe("ColdcardExportPublicKey", () => {
...
+  });
   it("has a message about exporting xpub", () => {
     expect(interactionBuilder({
       network: TESTNET,
       bip32Path: COLDCARD_BASE_BIP32,
     }).hasMessagesFor({
       state: PENDING,
       level: INFO,
-      code: "coldcard.export",
+      code: "coldcard.export_xpub",
```
bucko13, line 155: Yeah, I like these more descriptive codes 👍

src/coldcard.test.js
```diff
@@ -264,6 +296,12 @@ describe("ColdcardSignMultisigTransaction", () => {
       psbt,
     });
   }
+  describe("constructor", () => {
+    it("fails when sending in no psbt", () => {
+      expect(() => interactionBuilder({})).toThrow(/PSBT must be included/i);
+   });
+
```
bucko13, line 303: extra line.

src/coldcard.test.js
```diff
@@ -14,43 +13,50 @@ import {
...
+      expect(() => interactionBuilder({
+        bip32Path: 4,
+        network: TESTNET,
+      })).toThrow(/Bip32 path should be a string/i);
+    });
   });
 
-  describe('parse', () => {
+  describe("parse", () => {
+
```
bucko13, line 38: extra line

src/coldcard.js
humanumbrella: I agree with you, but eslint does not. 😬

humanumbrella: Oh interesting -- if you do 

```
if (!psbt) {
 throw new Error('...');
}
...
```

That's okay but if you do

```
if (!psbt) {
 throw new Error('...');
} else {
  ...
}
```

It shows as an eslint error. I just took it to mean can't do any negated condition, but apparently that's incorrect.

src/coldcard.js
humanumbrella: Good catch.

src/coldcard.js
bucko13: Oh yeah, that's probably like with `return`s. I forget what the rule is but I think I saw it in our other linting PR. Basically the else is unnecessary if you throw or return since if the condition is true, it'll never get past anyway and there's no need for a guard statement.

bucko13 (dismissed):
GTG!

src/fixtures.js
```diff
@@ -1,4 +1,4 @@
-import {COLDCARD_BASE_BIP32, WALLET_CONFIG_VERSION} from './coldcard';
```
dhruvbansal, line 1: Use shared fixtures from `unchained-bitcoin`.  Where not possible (e.g. Coldcard config format), store it in `coldcard.fixtures.js`.

dhruvbansal (changes requested):
Minor feedback on bip32 validation.  Use shared fixtures.

src/coldcard.js
```diff
@@ -56,21 +75,71 @@ export class ColdcardFileParser extends ColdcardInteraction {
     } else {
       throw new Error("Unknown network.");
     }
+    this.addressType = 'Unknown';
+    this.relativePath = '';
+    this.bip32ValidationErrors = [];
   }
 
+  // TODO make these messages more robust
+  //   (e.g use `menuchoices` as an array of `menuchoicemessages`)
```
dhruvbansal, line 84: Worth doing!

dhruvbansal (changes requested):

src/coldcard.js
```diff
@@ -91,36 +105,32 @@ export class ColdcardFileParser extends ColdcardInteraction {
...
-    //   (or at least a hardened path deeper than m/45')
+    //   (or at least a p2sh hardened path deeper than m/45')
 
     if (typeof file === "object") {
-      this.xpubJSONFile = file;
+      this.data = file;
     } else if (typeof file === "string") {
       try {
-        this.xpubJSONFile = JSON.parse(file);
+        this.data = JSON.parse(file);
```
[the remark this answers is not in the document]
humanumbrella: not keeping this as a class property at all anymore

humanumbrella:
NOTE:  Tests won't pass until we do a release and bump version of `unchained-bitcoin` here.

src/coldcard.js
humanumbrella: I will come back to this if I can.

humanumbrella:
I also added bip32 error handling via messages for Ledger and Trezor and appropriate unit tests here
Tests still won't pass til `unchained-bitcoin` is released and I bump the dep here. we can do that dance tomorrow.

dhruvbansal (approved):

waldenraines (approved):
This has already been approved by Dhruv prior to the version of `unchained-bitcoin` being updated.

PR #44, opened 2020-10-05
WIP: generalize multisig wallet config interface

bucko13:
An initial pass to see what the config interface could look like to be generalized to support multiple different config formats. This still only supports ColdCard but you should be able to see how more formats should be easily added on. 

I also split out some of the validation and instantiations. I think this makes the API a little more clear and also the unit testing a little more fine grained (though maybe this isn't desired). 

TODO:

- [x] Move existing functionality into its own config.js

- [x] Update tests for existing functionality

- [ ] Replace existing functionality from coldcard.js

- [ ] Add explicit caravan interfaces

- [ ] Add interface to convert _from_ a ColdCard config

- [ ] Update jsdoc strings


src/config.js
```diff
@@ -0,0 +1,391 @@
...
+          options.quorum = {
+            requiredSigners: m,
+            totalSigners: n,
+          } 
+          break;
+        }
+        default: {
+          // handle possible xfp: xpub pairs from config
+          try {
+            // will throw if not a valid
```
humanumbrella, line 259: forgot a i think - xpub

src/config.js
```diff
@@ -0,0 +1,391 @@
...
+            // will throw if not a valid 
+            ExtendedPublicKey.fromBase58(value);
+            
+            // check if the key is a valid fingerprint
+            validateXFP(key);
+
+            // get the network from xpub prefix
+            const prefix = value.slice(0, 4)
+            let network;
+            if (prefix === 'xpub') {
```
humanumbrella, line 268: prefix can be in x/Y/Zpub for mainnet and t/U/Vpub for testnet

src/config.js
```diff
@@ -0,0 +1,391 @@
...
+        // formatting
+        .map(line => line.trim())
+        // ignore comments
+        .filter(line => line[0] !== '#' && line.length);
+
+    // for each line separate key and values by colon
+    lines.forEach(line => {
+      // trim white space and set to lowercase
+      const [key, value] = line.split(':').map(str => str.trim());
+      // handle recognized keys (name, format, derivation, and policy)
```
humanumbrella, line 236: maybe not needed atm but derivation isn't handled

humanumbrella: we don't write this out bc right now it's a fixed path for all xpubs, which isn't necessarily true in our use case.

src/config.js
```diff
@@ -0,0 +1,391 @@
...
+   * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+   * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+   * 
+   * @param {string} configString valid coldcard config string
+   * @returns {MultisigWalletConfig} new instance of MultisigWalletConfig
+   * @example
+   * import { MultisigWalletConfig } from "unchained-wallets";
+   * 
+   * const config = MultisigWalletConfig.fromColdCardConfig(ccFileData);
+   * console.log(JSON.parse(config.toJSON()))
```
humanumbrella, line 185: I think this is toJSONConfig ? And in this example - it doesnt return startingAddress or client because they're unknown, right?

src/config.test.js
```diff
@@ -0,0 +1,138 @@
...
+    const definitelyNotJSON = 77;
+    const jsonConfigBad = { 'test': 0 };
+
+    expect(() => configBuilderFromJson(notJSON).toThrow(/Unable to parse JSON/i));
+    expect(() => configBuilderFromJson(definitelyNotJSON).toThrow(/Not valid JSON/i));
+    expect(() => configBuilderFromJson({}).toThrow(/Not valid JSON/i));
+    expect(() => configBuilderFromJson(jsonConfigBad).toThrow());
+  })
+  
+  it('should validate xpubs', () => {
```
humanumbrella, line 49: ?

src/config.test.js
```diff
@@ -0,0 +1,138 @@
...
+    expect(config.network).toEqual(options.network)
+    config.extendedPublicKeys.forEach((key) => {
+      const expectedKeys = options.extendedPublicKeys;
+      // config's xpub array should have the right xfp and xpub 
+      expect(
+        expectedKeys.findIndex(({ xfp, xpub }) => key.xfp === xfp && key.xpub === xpub)
+      ).toBeGreaterThan(-1)
+    })
+  })
+})
```
humanumbrella, line 138: space

humanumbrella (changes requested):
missing some ; at the end of lines - not sure if we care.

couple questions and some nits - we can discuss soon

src/config.js
```diff
@@ -0,0 +1,391 @@
...
+    
+    // TODO: how do we handle caravan configs where the extended public keys
+    // don't include xfp information? We can get a parent fingerprint 
+    // from the xpub and use that. It wouldn't be valid as an xfp but could be used as stand in? 
+    return new this(options);
+  }
+
+  /**
+   * @description static method wrapper around getting a new
+   * config based on a JSON since Caravan is the assumed format of accepted JSON
```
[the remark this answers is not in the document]
bucko13: I think I got this from the original file/code. There were some comments about "importing caravan config" or something like that.

src/config.js
```diff
@@ -0,0 +1,391 @@
...
+   * @param {string} [options.uuid] - used for name of config, will fallback to name option
+   * @param {string} [options.name] - required if no uuid passed
+   * @param {object} options.quorum - describes the `m` and `n` of a multisig wallet
+   * @param {number} options.quorum.requiredSigners - number of required signers, i.e. `m`
+   * @param {number} options.quorum.totalSigners - number of total signers, i.e. `n`
+   * @param {string} options.addressType - what kind of address (e.g. p2sh, p2wsh)
+   * @param {object[]} options.extendedPublicKeys - array of extended pub key objects
+   * @param {string} [options.extendedPublicKeys.xfp] - master fingerprint
+   * @param {string} options.extendedPublicKeys.xpub - valid xpub
+   * @param {string} options.extendedPublicKeys.xpub - valid xpub
```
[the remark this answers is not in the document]
bucko13: copy pasta. Good catch.

src/config.js
```diff
@@ -0,0 +1,391 @@
...
+    if (typeof jsonString !== 'string') throw new TypeError('Must pass a valid JSON string');
+    let options;
+
+    try {
+      options = JSON.parse(jsonString);
+    } catch (error) {
+      throw new Error("Unable to parse JSON.");
+    }
+    
+    // TODO: how do we handle caravan configs where the extended public keys
```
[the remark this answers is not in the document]
bucko13: Yeah, exactly. Or really any JSON config we can get. Presumably not all config setups have or require xfps since you can configure a multisig wallet without it.

src/config.js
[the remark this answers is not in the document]
bucko13: Oh yeah, mis-read this a little bit. The way I was reading the implementation was that unchained format is caravan format and vice versa. 

Here's where I read it referred to as a caravan config. It was in the tests:

src/config.js
bucko13: Thanks. Will add a utility that should probably also go into uc-bitcoin I think

src/config.js
bucko13: Yeah that's probably why I originally skipped (and the fixtures didn't have) it but I think it should still be worth adding now with some basic path validation.

src/config.js
bucko13: Yeah, that's right. They don't exist on the CC config so they will be skipped (and they're not required in the constructor). 

I'm now wondering if `toJSONConfig` is the right name for the method or if `toJSON` is better. Let me know what you think, but updating the comment to reflect the name of the method as it is now (as per your comment).

bucko13:
@humanumbrella ready for another review. I tried to catch the semi-colons where I could. That should really be getting handled by the linter...

bucko13:
Noting some requested changes here:
```
 if (client) {
      assert(
        this.supportedClients.includes(client.type),
        `Client type "${client.type}" not supported`
      )
      this.client = client;
    }
```
instead of
```
if (client) {
      assert(
        this.supportedClients.includes(client),
        `Client type "${client}" not supported`
      )
      this.client = client;
    }
```

```
* @param {Object} [options.client] - private or public, really only valid for caravan
```
instead of
```
* @param {string} [options.client] - private or public, really only valid for caravan
```

src/config.js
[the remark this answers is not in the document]
bucko13: @humanumbrella any further comment on this or good to just leave as a TODO?

bucko13:
@humanumbrella feedback addressed.

dhruvbansal (changes requested):

src/index.js
```diff
@@ -394,9 +394,7 @@ export function ConfirmMultisigAddress({keystore, network, bip32Path, multisig,
 export function ConfigAdapter({ KEYSTORE, jsonConfig }) {
```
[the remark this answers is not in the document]
bucko13: `generateColdcardConfig` and `parseColdcardConfig` have been moved out of the config class now into their own functions in the `coldcard.js` file. I left this Adapter class as is since this was the existing interface and I didn't want to interfere with any other downstream dependencies, but for someone who wants to functionally interact with a coldcard config, that capacity now exists, which has also simplified the config struct/class considerably.

src/coldcard.js
```diff
@@ -507,107 +516,177 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
+        case 'derivation': {
+          const pathError = validateBIP32Path(value);
+          assert(!pathError.length, pathError);
+          options.derivation = value;
+          break;
         }
-        if (typeof xpub.xfp !== 'string') {
-          throw new Error("XFP not a string");
+        case 'policy': {
+          let [m, n] = value.split(' of ');
```
humanumbrella, line 613: this can be a problem because coldcard supports many varieties of representing this

see https://github.com/Coldcard/firmware/blob/master/shared/multisig.py#L580

src/coldcard.js
```diff
@@ -507,107 +516,177 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
+   *        xpub: 'tpubD9ArfXowvGHnuECKdGXVKDM...FxPa',
+   *      },
+   *      {
+   *        xfp: '4369050F',
+   *        xpub: 'tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n',
+   *      }
+   *    ]
+   * }
+   */
+export function parseColdcardConfig(configString) {
```
humanumbrella, line 577: i didnt compare line by line but this would be good to compare how they do it vs how you've done it
https://github.com/Coldcard/firmware/blob/master/shared/multisig.py#L523

src/coldcard.js
```diff
@@ -507,107 +516,177 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
-Name: ${this.name}
-Policy: ${this.requiredSigners} of ${this.totalSigners}
-Format: ${this.addressType}
+Name: ${config.name}
+Policy: ${config.requiredSigners} of ${config.totalSigners}
+Format: ${config.addressType}
 
 `;
     // We need to loop over xpubs and output `xfp: xpub` for each
-    let xpubs = this.extendedPublicKeys.map((xpub) => `${xpub.xfp}: ${xpub.xpub}`);
```
humanumbrella, line 608: obviously will need an update after #49 and firmware version `3.2.1` from coldcard

src/coldcard.js
```diff
@@ -507,107 +516,177 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
-        if (xpub.xfp.length !== 8) {
-          throw new Error("XFP not length 8");
+        default: {
+          // handle possible xfp: xpub pairs from config
+          try {
+            // will throw if not valid base58 
+            ExtendedPublicKey.fromBase58(value);
+            
+            // check if the key is a valid fingerprint
+            validateRootFingerprint(key);
```
humanumbrella, line 627: i think it does make sense to make sure it's not in there twice rather than lettin em get all the way to the coldcard to find that out.  in caravan we could also edit it to not allow it

src/config.js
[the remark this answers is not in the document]
humanumbrella: My first thought was just start counting -- previous to `3.2.1` you could set em all to `0x00000000` but now that won't work - they need to be unique. So your suggestion of using the fingerprint that's in the xpub (even though it's not the root fingerprint) would probably be fine. 

If none of the xpubs have xfps - we should not do this though, because none of them will be able to sign (unless just happen to be at depth 1 and the parent fingerprint of the xpub *is* the root fingerprint.

src/config.js
```diff
@@ -0,0 +1,203 @@
...
+   */
+  static fromCaravanConfig(jsonString) {
+    return MultisigWalletConfig.fromJSON(jsonString);
+  }
+
+  /**
+   * @description Validate the public keys on the config instances. 
+   * Not all configs require xfp in the xpub objects, so an optional param is
+   * accepted to dictate whether or not this value should be validated
+   * @param {boolean} requiresXFP - does the object require an xfp, if so it will also be validated
```
humanumbrella, line 156: this is a little confusing because you could require a *single* xfp - and then the rest can be noise. but the way this works now is if you say requiresXFP every xpub must have an xfp.

src/fixtures.js
```diff
@@ -0,0 +1,154 @@
+import {COLDCARD_BASE_BIP32, WALLET_CONFIG_VERSION} from './coldcard';
```
humanumbrella, line 1: again some updates here

src/coldcard.js
```diff
@@ -507,107 +516,177 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
+   * BE26B07B: tpubD9ArfXowvGHnuECKdGXVKDM...FxPa
+   * 4369050F: tpubD8NXmKsmWp3a3DXhbihAYbY...9C8n
+   * 
+   * @param {string} configString valid coldcard config string
+   * @returns {MultisigWalletConfig} new instance of MultisigWalletConfig
+   * @example
+   * import { parseColdcardConfig } from "unchained-wallets";
+   * 
+   * const config = MultisigWalletConfig.fromColdCardConfig(ccFileData);
+   * console.log(JSON.parse(config.toJSONConfig()))
```
humanumbrella, line 547: this method name changed too I believe

humanumbrella (changes requested):
Just some small notes and changes requested (bringing in the changes from #49 will be important)

src/coldcard.js
```diff
@@ -507,107 +516,177 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
+   * 
+   * {
+   *    name: MeMyself,
+   *    quorum: {
```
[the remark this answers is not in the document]
bucko13: Yes, because this is how it is serialized in our json formatting/caravan. The options no longer have this but the json export makes it explicit.

src/coldcard.js
bucko13: Ah, good reference. Thanks!

src/coldcard.js
bucko13: To be clear, you're referencing https://github.com/Coldcard/firmware/blob/master/shared/multisig.py#L637 I'm guessing? It looks like this should be a regex instead then, just splitting up two digits.

src/coldcard.js
bucko13: I agree. I think that's one nice thing about having the individual wallet client config methods kind of separate is that it makes sense to enforce the same things coldcard will enforce (as best we can).

src/coldcard.fixtures.js
```diff
@@ -130,5 +127,17 @@ Format: P2SH
...
-`
+`,
+  coldcardConfigDerivation: `# Coldcard Multisig setup file (exported from unchained-wallets)
+# https://github.com/unchained-capital/unchained-wallets
+# v${COLDCARD_WALLET_CONFIG_VERSION}
+# 
+Name: OWPyFOA1
+Policy: 2 of 3
+Format: P2SH
+Derivation: m/45'/1/0
```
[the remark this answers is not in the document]
bucko13: e.g.
deriv1
key1
deriv2
key2
deriv1 (again, repeat)
key3

src/config.js
bucko13: I would say that those two statements are equivalent. i.e. 1 valid xpfs + 2 "noisy" xfps == every xpub has an xfp. Basically what this is saying that if a root fingerprint is required, make sure it's included. We can't validate that it's not noisy, but that's not really relevant in these operations since we can't really validate them anyway. So if it's required, make sure they all have _something_ and that it's valid hex, correct length, etc.

src/coldcard.js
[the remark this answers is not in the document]
bucko13: Though looking at the fixtures, there's some inconsistency here which I think you're pointing out. I'll try and reconcile this. But basically, we have the "options" format and we have the json/caravan config format. Prior to recent reviews of this PR they were one and the same, but now that's not as true as it once was.

src/config.js
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * @returns {string} returns a JSON string with config values
+   */
+  toJSON() {
+    return JSON.stringify({
+      name: this.name,
+      addressType: this.addressType,
+      network: this.network,
+      quorum: this.quorum,
```
bucko13, line 197: Note: This needs to be fixed.`quorum` is no longer in the MultisigWalletConfig properties, but is in the Caravan config. Needs to be inferred from other data for `CaravanConfig` and here should be replaced with `requiredSigners`

src/config.js
```diff
@@ -0,0 +1,203 @@
...
+      startingAddressIndex: this.startingAddressIndex,
+      extendedPublicKeys: this.extendedPublicKeys,
```
bucko13, line 199: This will only work if extendedPublicKeys are strings. Need to either catch situation where it's not or force it to be strings.

src/config.js
```diff
@@ -0,0 +1,203 @@
...
+      }
+
+      return true
+    })
+  }
+
+  /**
+   * @description A simple wrapper for returning a json config
+   * since the params of the class are valid for caravan.
+   * @returns {MultisigWalletConfig} a new MultisigWalletConfig instance
```
bucko13, line 183: Doesn't return `MultisigWalletConfig`

src/config.js
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  /**
+   * @description static method wrapper around getting a new
+   * config based on a JSON since Caravan is the assumed format of accepted JSON
+   * @param {string} jsonString - valid config json string from Caravan export
+   * @returns {MultisigWalletConfig} new instance of a MultisigWalletConfig
+   */
+  static fromCaravanConfig(jsonString) {
+    return MultisigWalletConfig.fromJSON(jsonString);
```
bucko13, line 149: See comment about difference between JSON and caravan config now.

src/config.js
```diff
@@ -0,0 +1,203 @@
...
+   * @param {object} options - options object to create config from
+   * @param {string} [options.uuid] - used for name of config, will fallback to name option
+   * @param {string} [options.name] - required if no uuid passed
+   * @param {string} [options.derivation] - BIP32 path for xpub derivation
+   * @param {number} options.requiredSigners - number of required signers, i.e. `m`
+   * @param {string} options.addressType - what kind of address (e.g. p2sh, p2wsh)
+   * @param {object[]} options.extendedPublicKeys - array of extended pub key objects
+   * @param {string} [options.extendedPublicKeys.xfp] - master fingerprint
+   * @param {string} options.extendedPublicKeys.xpub - valid xpub
+   * @param {number} [options.startingAddressIndex] - an optional positive integer indicating
```
bucko13, line 49: re-order. This is in the middle of the xpub options.

src/config.js
bucko13: I'm addressing this with an `addPlaceholderFingerprint` method on the class that can be used when we have at least one xfp (assumed valid) and need the rest to have a placeholder in there.

src/coldcard.js
bucko13: Added new check in `validateExtendedPublicKeys` which gets called when instantiating the config.

src/index.js
dhruvbansal, line 394: `ConfigAdapter` -> `MultisigWalletConfig`
`KEYSTORE` -> `keystore` XOR `coordinator` or something like `target` (?)
`jsonConfig` -> `braids` or `braidSet`
add `options`

dhruvbansal (changes requested):

PR #46, opened 2020-12-10
Properly close Ledger transports after an interaction runs 

humanumbrella:
(or errors…). Not doing this previously was preventing using the Ledger in multiple tabs. (E.g. `Error: Unable to claim interface.`)

Test suite run screenshot attached. With this code, each interaction you have with the device will generate the USB logo in the tab in Chrome. 

Screen Shot 2020-12-10 at 2 49 07 PM

But instead of this persisting after the interaction, the USB logo should fade away on success or failure. The next interaction re-creates the transport.

Screen Shot 2020-12-10 at 2 41 51 PM

### DIFF
--- a/src/ledger.js
+++ b/src/ledger.js
@@ -224,6 +224,20 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
     });
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


bucko13 (approved):
Nice addition. LGTM

Issue #48, opened 2020-12-15
Replace disallowed cold card characters in wallet names with permissible ones


waldenraines:
Unfortunately cold card truncates the names of multisignature wallets that contain characters it doesn't like.  

We should replace spaces and other "invalid" characters with valid ones so you don't end up with a situation like this:

![img_20201214_142934583](https://user-images.githubusercontent.com/4116405/102157723-ffcd9880-3e4d-11eb-9347-f532dd8f7e46.jpg)

humanumbrella:
This problem is a bug in Coldcard wallet config processing code ... they use `#` as a comment character and truncate the line up to the `#` - but that should only happen if the `#` is the first character on the line.

See -> https://github.com/Coldcard/firmware/blob/master/shared/multisig.py#L543-L549

humanumbrella:
They put a fix in for this one with latest changes after I informed

See -> https://github.com/Coldcard/firmware/blob/master/shared/multisig.py#L555

PR #49, opened 2021-01-09
Quickfix: update coldcard wallet config output to include bip32paths

humanumbrella:
Coldcard released firmware [version 3.2.1](https://github.com/Coldcard/firmware/tree/2021-01-07T1439-v3.2.1) today. It has several multisig updates.

Originally we left out the `Derivation` parameter in our wallet config files (even though we had (and still have) access to it via what comes in) because the Coldcard was only paying attention to the first entry for `Derivation`. And in an effort to alleviate confusion (and because it wasn't a required parameter at the time), we left it out completely and happily continued with `Derivation: Unknown`.

Xpub specific derivations are now supported and `Derivation: Unknown` is no longer supported. Added logic to mask out the derivation to the proper depth (included in the xpub itself) if we do not know the bip32path for a particular xpub in the braid.

Another major change in the latest firmware update is that multiple xpubs from the same xfp are no longer supported. We had a fixture setup like this, so I updated the fixture.

More line changes than expected because I ran it through prettier.

I expect that we will be integrating changes from https://github.com/unchained-capital/unchained-wallets/pull/44 soon, but I wanted to go ahead and submit a solution to the current situation. I believe this does it. If you use the older firmware, it will only look at the first derivation path, but I believe that will be okay.

### DIFF
--- a/src/coldcard.js
+++ b/src/coldcard.js
@@ -29,33 +29,27 @@ import {
   INFO,
   ERROR,
 } from "./interaction";
-import {
-  P2SH,
-  P2SH_P2WSH,
-  P2WSH,
-} from 'unchained-bitcoin';
+import { P2SH, P2SH_P2WSH, P2WSH } from "unchained-bitcoin";
 
-export const COLDCARD = 'coldcard';
+export const COLDCARD = "coldcard";
 // Our constants use 'P2SH-P2WSH', their file uses 'P2SH_P2WSH' :\
 export const COLDCARD_BASE_BIP32_PATHS = {
   "m/45'": P2SH,
-  "m/48'/0'/0'/1'": P2SH_P2WSH.replace('-', '_'),
+  "m/48'/0'/0'/1'": P2SH_P2WSH.replace("-", "_"),
   "m/48'/0'/0'/2'": P2WSH,
-  "m/48'/1'/0'/1'": P2SH_P2WSH.replace('-', '_'),
+  "m/48'/1'/0'/1'": P2SH_P2WSH.replace("-", "_"),
   "m/48'/1'/0'/2'": P2WSH,
 };
 const COLDCARD_BASE_CHROOTS = Object.keys(COLDCARD_BASE_BIP32_PATHS);
 
-export const COLDCARD_WALLET_CONFIG_VERSION = "0.0.1";
-
+export const COLDCARD_WALLET_CONFIG_VERSION = "0.0.2";
 
 /**
  * Base class for interactions with Coldcard
  *
  * @extends {module:interaction.IndirectKeystoreInteraction}
  */
-export class ColdcardInteraction extends IndirectKeystoreInteraction {
-}
+export class ColdcardInteraction extends IndirectKeystoreInteraction {}
 
 /**
  * Base class for JSON Multisig file-based interactions with Coldcard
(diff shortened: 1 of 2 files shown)


src/coldcard.js
```diff
@@ -29,33 +29,27 @@ import {
...
   "m/48'/0'/0'/2'": P2WSH,
-  "m/48'/1'/0'/1'": P2SH_P2WSH.replace('-', '_'),
+  "m/48'/1'/0'/1'": P2SH_P2WSH.replace("-", "_"),
   "m/48'/1'/0'/2'": P2WSH,
 };
 const COLDCARD_BASE_CHROOTS = Object.keys(COLDCARD_BASE_BIP32_PATHS);
 
-export const COLDCARD_WALLET_CONFIG_VERSION = "0.0.1";
-
+export const COLDCARD_WALLET_CONFIG_VERSION = "0.0.2";
```
bucko13, line 45: Does this version come from coldcard or is it our own internal tracking?

humanumbrella: internal tracking

bucko13: Got it. I like this idea in general, of versioning config standards. Depending on what direction we go in with #44 this might be something to consider tracking for all configs we support (including caravan's JSON). 

Wonder what you think of making this a bigger bump then a "patch"? I understand this is kind of arbitrary, but given that these two versions are basically not compatible, might be worth making it `0.1.0` or even `1.0.0`? I'm not super opinionated either way, but thought I'd throw it out there.

waldenraines:
> If you use the older firmware, it will only look at the first derivation path, but I believe that will be okay.

I'll test the prior version of the firmware, have to downgrade anyway due to a bug in cold card.

src/coldcard.js
```diff
@@ -593,8 +609,6 @@ export class ColdcardMultisigWalletConfig {
    * @returns {string} output to be written to a text file and uploaded to Coldcard.
    */
   adapt() {
-    // Coldcard configs can't have spaces in the names, it just splits on space and takes the first word.
```
bucko13, line 596: Is this no longer true about about names?

src/coldcard.js
```diff
@@ -604,10 +618,17 @@ Policy: ${this.requiredSigners} of ${this.totalSigners}
...
-    // We need to loop over xpubs and output `xfp: xpub` for each
-    let xpubs = this.extendedPublicKeys.map((xpub) => `${xpub.xfp}: ${xpub.xpub}`);
-    output += xpubs.join("\r\n");
-    output += "\r\n";
+    // We need to loop over xpubs and output `Derivation: bip32path` and `xfp: xpub` for each
+    let xpubs = this.extendedPublicKeys.map((xpub) => {
+      // Mask the derivation to the appropriate depth if it is not known
+      const unknownBip32 = xpub.bip32Path.toLowerCase().includes("unknown");
+      const derivation = unknownBip32
+        ? `m${"/0".repeat(ExtendedPublicKey.fromBase58(xpub.xpub).depth)}`
```
bucko13, line 626: Neat!

bucko13 (commented):
LGTM! Sounds like we're leaning towards updating the version number to imply this isn't backwards compatible.

src/coldcard.js
humanumbrella: Ya - turns out it wasn't spaces. the problem was that our default names were e.g. `Vault #1` and then the # would cause the line to be ignored making it seem like the space was the problem.

See https://github.com/unchained-capital/unchained-wallets/issues/48 for more details

src/coldcard.js
humanumbrella: changing it to `1.0.0`

bucko13 (approved):

waldenraines:
Old firmware, old config - Works
Old firmware, new config - Works
New firmware, old config - Doesn't work
New firmware, new config - Works

src/coldcard.js
waldenraines: Yeah, they fixed that in the latest firmware version

waldenraines (approved):
Everything I tested worked as expected

PR #50, opened 2021-02-12
Adds "custom" text-based device and re-orders keystores alphabetically

humanumbrella:
We need to support custom flows for text-based xpub upload and custom PSBT signing, so I have created the interaction classes below to support these use cases.

Includes unit tests and fixture data.

Screen Shot 2021-02-12 at 12 58 15 PM

I have also edited the exports and deciding device interaction class such that they're in alphabetic order. This will cause conflicts when merging into other existing PRs, but as we add more keystores, this will be a nice thing to keep in mind.

If we think it will cause too much hassle, lmk and I can remove it.

### DIFF
--- a/src/custom.fixtures.js
+++ b/src/custom.fixtures.js
@@ -1,0 +1,34 @@
+import { TEST_FIXTURES, ROOT_FINGERPRINT } from "unchained-bitcoin";
+
+const { nodes } = TEST_FIXTURES.keys.open_source;
+
+const P2SH_BASE_MAIN = "m/45'/0'/0'";
+const P2SH_BASE_TEST = "m/45'/1'/0'";
+
+export const customFixtures = {
+  // These use the Open Source Wallet words from the `unchained-bitcoin` fixtures
+  validCustomTpubJSON: {
+    bip32Path: P2SH_BASE_TEST,
+    xpub: nodes[P2SH_BASE_TEST].xpub,
+    rootFingerprint: ROOT_FINGERPRINT,
+  },
+  validCustomXpubJSON: {
+    bip32Path: P2SH_BASE_MAIN,
+    xpub: nodes[P2SH_BASE_MAIN].xpub,
+    rootFingerprint: ROOT_FINGERPRINT,
+  },
+
+  validTpubFakeRootFingerprintOutput: {
+    xpub:
+      "tpubDDQubdBx9cbs16zUhpiM135EpvjSbVz7SGJyGg4rvRVEYdncZy3Kzjg6NjuFWcShiCyNqviWTBiZPb25p4WcaLppVmAuiPMrkR1kahNoioL",
+    bip32Path: P2SH_BASE_TEST,
+    rootFingerprint: "0b287198",
+  },
+
+  validXpubFakeRootFingerprintOutput: {
+    xpub:
+      "xpub6CCHViYn5VzKFqrKjAzSSqP8XXSU5fEC6ZYSncX5pvSKoRLrPDcF8cEaZkrQvvnuwRUXeKVjoGmAqvbwVkNBFLaRiqcdVhWPyuShUrbcZsv",
+    bip32Path: P2SH_BASE_MAIN,
+    rootFingerprint: "266afe03",
+  },
+};
--- a/src/custom.js
+++ b/src/custom.js
@@ -1,0 +1,272 @@
+/**
+ * Provides classes for interacting via text-based copy/paste XPUBs and
+ * download/sign generic PSBT files using a custom "device'
+ *
+ * The following API classes are implemented:
+ *
+ * * CustomExportExtendedPublicKey
+ * * CustomSignMultisigTransaction
+ *
+ * @module custom
+ */
+import {
+  unsignedMultisigPSBT,
+  parseSignaturesFromPSBT,
+  MAINNET,
+  TESTNET,
+  validateBIP32Path,
+  validateRootFingerprint,
+  ExtendedPublicKey,
+} from "unchained-bitcoin";
+import {
+  IndirectKeystoreInteraction,
+  PENDING,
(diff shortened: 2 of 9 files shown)


humanumbrella:
woof, OK I think I got it. `unchained-bitcoin` hadn't been bumped in a while.

src/custom.fixtures.js
```diff
@@ -0,0 +1,34 @@
+import { TEST_FIXTURES, ROOT_FINGERPRINT } from "unchained-bitcoin";
```
bucko13, line 1: I was starting to do this in the config branch, but what do you think about having a fixtures (or sometimes I've seen it named `data`) folder? I imagine there will only be more of these as time goes on and could help keep things organized.

humanumbrella: I could go either way, I think given we already have
`name`
`name.test`

adding in
`name.fixtures` seems fine.

but also given the scale this could reach, then the TLD gets pretty crowded so could also see 
`data`
`data/name.fixtures` as well.

are you advocating only moving fixtures, or also trying to consolidate where tests live?

bucko13: Only moving fixtures

src/custom.test.js
```diff
@@ -0,0 +1,263 @@
...
+      const missingXpub = {
+        ...customFixtures.validCustomTpubJSON,
+      };
+      Reflect.deleteProperty(missingXpub, "xpub");
+      expect(() => interaction.parse(missingXpub)).toThrow(
+        /Not a valid ExtendedPublicKey/i
+      );
+    });
+
+    it("testnet no rootFingerprint is included so it gets computed", () => {
```
bucko13, line 74: some of the wording of the tests feel a little weird to me. I like to have it read as if it works in a sentence, e.g. the previous one `it fails when missing xpub field`. Not all of them are like that though. This is partially just personal preference, but I think it would be good to be consistent.

src/custom.test.js
```diff
@@ -0,0 +1,263 @@
...
+      const interaction = interactionBuilder({
+        network: TESTNET,
+        psbt: multisigs[0].psbt,
+      });
+      const result = interaction.request();
+      expect(result).toEqual(multisigs[0].psbt);
+    });
+
+    it("construct psbt if there is not one", () => {
+      const interaction = interactionBuilder({
```
bucko13, line 181: might be a little cleaner to pull out `transaction = transactions[0]`.

src/custom.test.js
```diff
@@ -0,0 +1,263 @@
...
+      });
+      const result = interaction.request().toBase64();
+      expect(result).toEqual(transactions[0].psbt);
+    });
+  });
+
+  describe("parse", () => {
+    it("return multi input, single signature set", () => {
+      const interaction = interactionBuilder({
+        psbt: multisigs[0].psbtPartiallySigned,
```
bucko13, line 195: same here for `multisigs[0]`

bucko13 (changes requested):

src/index.js
```diff
@@ -1,13 +1,23 @@
-import {version} from "../package.json";
```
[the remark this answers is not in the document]
humanumbrella: should be **only** that in this file! hehe, check me if it's not though

src/custom.js
```diff
@@ -0,0 +1,272 @@
...
+      throw new Error(
+        `Root fingerprint validation error: ${e.message.toLowerCase()}.`
+      );
+    }
+    const numSlashes = this.bip32Path.split("/").length;
+    const bipDepth = this.bip32Path.startsWith("m/")
+      ? numSlashes - 1
+      : numSlashes;
+
+    if (xpubClass.depth !== bipDepth) {
```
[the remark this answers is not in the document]
humanumbrella: Yes in basically the same way as any other interaction. 
The bip32path can't be determined from the xpub itself, it only has depth embedded inside. 

This is just a check that what the caller *said* the bip32path was **at least** matches the depth of what you provided because that's as much as we can check.

E.g. if you said m/45' and gave me m/45'/0'/0' -- we should say `hold on a minute...`

src/custom.js
```diff
@@ -0,0 +1,272 @@
...
+    let rootFingerprint;
+    try {
+      xpubClass = ExtendedPublicKey.fromBase58(data.xpub);
+    } catch (e) {
+      throw new Error("Not a valid ExtendedPublicKey.");
+    }
+    try {
+      const pkLen = xpubClass.pubkey.length;
+      // If no fingerprint is provided, we will assign one deterministically
+      const fakeRootFingerprint = xpubClass.pubkey.substring(pkLen - 8);
```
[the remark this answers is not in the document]
humanumbrella: ya, sure that's a little cleaner.

src/custom.js
```diff
@@ -0,0 +1,272 @@
...
+
+  isSupported() {
+    return this.validationErrorMessages.length === 0;
+  }
+
+  messages() {
+    const messages = super.messages();
+
+    if (this.validationErrorMessages.length) {
+      this.validationErrorMessages.map((e) => {
```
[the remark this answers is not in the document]
humanumbrella: oops...

src/index.js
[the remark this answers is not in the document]
bucko13: Yeah, it was as far as I could tell just wanted to confirm I wasn't missing something!

src/custom.js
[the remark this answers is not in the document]
bucko13: > The bip32path can't be determined from the xpub itself, it only has depth embedded inside.

Oh yeah, duh. Sorry.

src/custom.test.js
humanumbrella: Word, good point re: consistency - edited some of the names.

bucko13 (approved):
GTG!

PR #53, opened 2021-04-30
Update the export message to include account

waldenraines:
Update the message to include the requirement of entering 0 for
the account number.

### DIFF
--- a/src/coldcard.js
+++ b/src/coldcard.js
@@ -102,6 +102,12 @@ class ColdcardMultisigSettingsFileParser extends ColdcardInteraction {
     messages.push({
       state: PENDING,
       level: INFO,
+      code: "coldcard.select_account",
+      text: "Enter 0 for account",
+    });
+    messages.push({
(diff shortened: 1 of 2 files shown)


humanumbrella (approved):
While this new message is optional per the latest `4.1.*+` releases, I think it's okay to not include something like `(if shown)`.

If someone is on an older firmware version, after the first instruction, e.g. "Go to Settings > Multisig Wallets > Export XPUB" - it will just work so they probably won't look at the messages anymore.

If the firmware asks them a question about account they will probably then come back and scan the messages to see that we tell them to enter 0.

LGTM

humanumbrella (changes requested):
Actually - would you mind adding something [here](https://github.com/unchained-capital/unchained-wallets/blob/master/src/coldcard.js#L225) about if you're on `Additionally firmware version 4.1.0 is not supported, please update.`

humanumbrella (approved):
here we go!

PR #59, opened 2021-06-08
Peg Ledger WebUSB to version `5.53.0` until Chromium releases their patch

humanumbrella:
In order for us to release any updates to Caravan, we need to peg the ledger-webusb npm package to `0.53.0` because they reverted that fix in `0.53.1` and it's just broken and they're waiting on Chromium to release.

Things work fine for us on this patched version, so let's stay here until after Chromium releases a patch or Ledger figures out another fix.

The lock file already has this version, so no changes expected there.


waldenraines (approved):
LGTM

PR #61, opened 2021-08-18
Adds support for Ledger/Trezor PSBT signing for P2SH case

humanumbrella:
Utilizes new functionality in `unchained-bitcoin` to allow a Ledger to sign a PSBT, working in `P2SH` for now.

If the PSBT is defined when we call the interaction class, we will take the provided PSBT and translate it into
- inputs
- outputs
- bip32Paths

which can be sent to the Ledger for signing.

Again, if the PSBT is defined when we call the interaction class, once the signatures come back from the device, we will return a partially signed PSBT (with the signatures returned from the device inserted and validated) instead of the normal array of signatures.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -244,6 +244,8 @@ export function ExportExtendedPublicKey({
  * @param {object[]} options.inputs - transaction inputs
  * @param {object[]} options.outputs - transaction outputs
  * @param {string[]} options.bip32Paths - the BIP32 paths on this device corresponding to a public key in each input
+ * @param {string} [options.psbt] - the unsigned_psbt
+ * @param {object} [options.keyDetails] - Signing Key Fingerprint + Bip32 Root
  * @return {module:interaction.KeystoreInteraction} keystore-specific interaction instance
  * @example
  * import {
@@ -286,6 +288,7 @@ export function SignMultisigTransaction({
   outputs,
   bip32Paths,
   psbt,
+  keyDetails,
 }) {
   switch (keystore) {
     case COLDCARD:
@@ -316,6 +319,8 @@ export function SignMultisigTransaction({
         inputs,
         outputs,
         bip32Paths,
+        psbt,
+        keyDetails,
       });
     case TREZOR:
       return new TrezorSignMultisigTransaction({
--- a/src/index.js
+++ b/src/index.js
@@ -246,6 +246,7 @@ export function ExportExtendedPublicKey({
  * @param {string[]} options.bip32Paths - the BIP32 paths on this device corresponding to a public key in each input
  * @param {string} [options.psbt] - the unsigned_psbt
  * @param {object} [options.keyDetails] - Signing Key Fingerprint + Bip32 Root
+ * @param {boolean} [options.returnSignatureArray] - return an array of signatures instead of a signed PSBT (useful for test suite)
  * @return {module:interaction.KeystoreInteraction} keystore-specific interaction instance
  * @example
  * import {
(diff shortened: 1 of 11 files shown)


waldenraines:
See https://github.com/unchained-capital/unchained-bitcoin/pull/37

src/ledger.js
```diff
@@ -991,6 +1015,12 @@ export class LedgerSignMultisigTransaction extends LedgerBitcoinInteraction {
     return (transactionSignature || []).map(inputSignature => `${signatureNoSighashType(inputSignature)}01`);
   }
 
+  hexSignaturesToBuffers(transactionSignature) {
+    // Ledger signatures include the SIGHASH byte (0x01) if signing for P2SH-P2WSH or P2WSH ...
+    // but NOT for P2SH ... This function should always return the signature with SIGHASH byte appended.
+    return (transactionSignature || []).map(inputSignature => Buffer.from(`${signatureNoSighashType(inputSignature)}01`,"hex"));
```
waldenraines, line 1021: Seems like maybe we ought to DRY this up?

https://github.com/unchained-capital/unchained-wallets/blob/8a9e07f98545ed5b9f084b9f2adbad70e0a1da39/src/ledger.js#L989-L991

waldenraines (changes requested):
Minor changes requested.

> Again, if the PSBT is defined when we call the interaction class, once the signatures come back from the device, we will return a partially signed PSBT (with the signatures returned from the device inserted and validated) instead of the normal array of signatures.

So then how can I get a PSBT defined in order to test this in, say, caravan?

src/ledger.js
humanumbrella: Sure

src/ledger.js
```diff
@@ -978,17 +995,28 @@ export class LedgerSignMultisigTransaction extends LedgerBitcoinInteraction {
...
+        } else {
+          return this.parseSignature(transactionSignature, "hex");
+        }
       } finally {
         transport.close();
       }
     });
   }
 
-  parse(transactionSignature) {
```
humanumbrella, line 988: really shouldn't have even been using the `parse` method -- that's not supposed to be used on `DirectKeystoreInteraction` classes.

humanumbrella (commented):
eslint fixed a few other files so this PR touches more files than just ledger

humanumbrella:
> Minor changes requested.
> 
> > Again, if the PSBT is defined when we call the interaction class, once the signatures come back from the device, we will return a partially signed PSBT (with the signatures returned from the device inserted and validated) instead of the normal array of signatures.
> 
> So then how can I get a PSBT defined in order to test this in, say, caravan?

There is no exposed UI for this at the moment, so while any PSBT would do, it'd have to be tested in something that's utilizing the library directly, e.g. a feature branch of another repo.

The unit tests give an example using the fixtures and further testing in `unchained-bitcoin` ... when we do expose this to a UI then we can test individual examples.

humanumbrella:
I did run through the ledger test suite since I refactored a little bit. btc app `1.6.1` firmware `2.0.0`
it all worked for me.

src/index.js
```diff
@@ -286,6 +288,7 @@ export function SignMultisigTransaction({
   outputs,
   bip32Paths,
   psbt,
+  keyDetails,
```
dhruvbansal, line 291: `signingKeyDetails` would be a better name.

dhruvbansal (changes requested):
One minor change requested -- but let's hold off on merging this till we update the Trezor part.

dhruvbansal (approved):

humanumbrella:
for testing https://humanumbrella.github.io/caravan/#/test (will PR up to public caravan shortly)

waldenraines (approved):

PR #66, opened 2021-12-17
Bugfix - missed a function call replace in PR#61

humanumbrella:
In https://github.com/unchained-capital/unchained-wallets/pull/61 - we removed the references to `parse` as they should not be on `DirectKeystoreInteraction`s ... forgot to replace the ones in Pubkey / Xpub Export -- this fixes that problem.

### DIFF
--- a/src/trezor.js
+++ b/src/trezor.js
@@ -186,7 +186,7 @@ try {
  *     ];
  *   }
  *
- *   parse(payload) {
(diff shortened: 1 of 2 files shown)


humanumbrella:
Screen Shot 2021-12-17 at 10 23 40 AM
and tested confirm on device as well.

dhruvbansal (approved):

PR #68, opened 2022-02-21
Support trezor signing on regtest

humanumbrella:


src/trezor.js
```diff
@@ -105,12 +105,16 @@ export const TREZOR_BOTH_BUTTONS = 'trezor_both_buttons';
  */
 export const TREZOR_PUSH_AND_HOLD_BUTTON = 'trezor_push_and_hold_button';
 
+const TREZOR_CONNECT_URL = "https://localhost:8088/";
+const TREZOR_BLOCKBOOK_URL = "http://localhost:3035"
```
bucko13, line 109: can we put these in process env too and have these values as fallbacks?

src/trezor.js
```diff
@@ -123,9 +127,7 @@ try {
 } catch(e) {
   // We hit this if we run this code outside of a browser, for example
   // during unit testing.
-  /* eslint-disable no-process-env */
-  if (process.env.NODE_ENV !== 'test') {
-    /* eslint-enable */
+  if (env_variables.NODE_ENV !== 'test') {
```
bucko13, line 130: I wonder if this will actually quiet the below error in testing now. Could be that it wasn't working because of the direct access issue in the comment above.

humanumbrella: I only noticed the direct access issue when trying to import this into react, otherwise it works fine (also tests pass, w00t!)

src/trezor.js
humanumbrella: good call

src/trezor.js
humanumbrella: Oops my previous comment was a quick reply with other stuff going on, so didn't quite get what you were saying.

Ya - tests pass now and I don't see any error messages. 
See before:

Screen Shot 2022-02-24 at 12 11 17 AM

and after this change:

Screen Shot 2022-02-24 at 12 11 46 AM

bucko13: fantastic. good snipe

bucko13 (approved):
This looks good to me. I say we merge and release to simplify development.

humanumbrella:
I think early next week will probably work best

PR #69, opened 2022-02-28
AP-2604 Fixing confirm-on-device for trezor.

billytetrud:
https://unchained-capital.atlassian.net/browse/AP-2604


bucko13 (approved):
Nice. and we got tests.

PR #81, opened 2022-11-08
optional publish step

steveblackmon:
add capability for jenkins to publish builds


nk1tz:
closing until further notice

PR #88, opened 2023-02-11
Ledgerv2 app version detection

bucko13:
This could still use some cleanup I think but it gets the basics done. A lot going on in this PR, commit history should give a good idea of the changes but summary: 

- more typescript fixes
- prettier formatting as I went into some more files
- adds a new vendor "dependency" for ledger-bitcoin support. Unfortunately the new js client is published anywhere. Since we're already using a vendor pattern for similar dependencies, I opted to do the same here. See the src/vendor/README.md in this PR for more details
- the vendor modules needed to be moved into src/ for proper transpiling
- some configs and tests needed to be updated after the above vendor changes
- ledger.ts updates to be able to export xpubs with the new app (tested as working in Caravan test suite! needs some [uc-bitcoin updates](https://github.com/unchained-capital/unchained-bitcoin/pull/58) to work)
- adds a couple catches to unsupported methods. I'm not convinced I've done the cleanest approach (not at all DRY as is) but the OPP as we have it is a little hectic so might be hard to generalize any further. Open to suggestions though!  

One thing that we might want to build out later is changing the messages that are pushed on the stack for each interaction, being conditional on the version of the app. 

NOTE: this PR isn't as big as it looks since there are a lot of "files changed" that are just pulling in vendor files from ledger. (and moving the existing vendors files into the src to make available for build, not sure if this is the best approach so open to feedback there). Most important changes are in ledger.ts.

Demo:

https://user-images.githubusercontent.com/4344978/218594955-5fffe758-f018-4697-8db4-f732aeecc468.mov


src/ledger.ts
```diff
@@ -758,25 +861,34 @@ export class LedgerExportExtendedPublicKey extends LedgerExportHDNode {
...
+        const xpub = deriveExtendedPublicKey(
+          this.bip32Path,
+          walletPublicKey.publicKey,
+          walletPublicKey.chainCode,
+          +fingerprint,
+          this.network
+        );
+
+        if (this.includeXFP) {
+          let rootFingerprint = await this.getXfp();
```
tylerchambers, line 876: const?

src/vendor/ledger-bitcoin/src/appClient.ts
```diff
@@ -0,0 +1,319 @@
...
+    const pathElements = pathStringToArray(path);
+
+    const clientInterpreter = new ClientCommandInterpreter();
+
+    // prepare ClientCommandInterpreter
+    const nChunks = Math.ceil(message.length / 64);
+    const chunks: Buffer[] = [];
+    for (let i = 0; i < nChunks; i++) {
+      chunks.push(message.subarray(64 * i, 64 * i + 64));
+    }
```
tylerchambers, line 300: how do we feel about: 

```typescript
  const chunkSize = 64;
  const chunks = [];
  for (let i = 0; i < message.length; i += chunkSize) {
    const chunk = message.slice(i, i + chunkSize);
    chunks.push(chunk);
  }
```

src/vendor/ledger-bitcoin/src/buffertools.ts
```diff
@@ -0,0 +1,134 @@
+import { createVarint, parseVarint, sanitizeBigintToNumber } from "./varint";
+
+export function unsafeTo64bitLE(n: number): Buffer {
+  // we want to represent the input as a 8-bytes array
+  if (n > Number.MAX_SAFE_INTEGER) {
```
tylerchambers, line 5: check if n < 0?

Also, Idk enough about typescript to know, but will we ever need to represent numbers > Number.MAX_SAFE_INTEGER ?

tylerchambers (approved):
Didn't realize I was reviewing code in the vendor directory until I was like halfway through. Some small nits but otherwise your part LGTM.

src/ledger.ts
```diff
@@ -204,6 +214,23 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
...
+      const response: AppAndVersion = await getAppAndVersion(transport);
+      this.appVersion = response.version;
+      return this.appVersion;
+    });
+  }
+
+  async isLegacyApp(): Promise<boolean> {
+    const version = await this.setAppVersion();
+    const [majorVersion, minorVersion] = version.split(".");
+    if (+majorVersion > 1 && +minorVersion >= 1) {
```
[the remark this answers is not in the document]
bucko13: I like it! Thanks [8a0047a](https://github.com/unchained-capital/unchained-wallets/pull/88/commits/8a0047a8e7bba16086de61062e2c4ac2491edb72)

src/ledger.ts
bucko13: [8a0047a](https://github.com/unchained-capital/unchained-wallets/pull/88/commits/8a0047a8e7bba16086de61062e2c4ac2491edb72)

src/vendor/ledger-bitcoin/src/appClient.ts
bucko13: I like it but... vendors ammiright? 😏

src/vendor/ledger-bitcoin/src/buffertools.ts
bucko13: tbh, they might take PRs if you wanted to suggest, but also i imagine you might enjoy taking a look at their rust code too if you were curious.

tylerchambers (approved):
LGTM!  (still)

src/vendor/bcur/index.js
```diff
@@ -1,2 +1,2 @@
-export { encodeUR } from './encodeUR';
-export { smartDecodeUR, decodeUR, extractSingleWorkload } from './decodeUR';
+export { encodeUR } from "./encodeUR";
+export { smartDecodeUR, decodeUR, extractSingleWorkload } from "./decodeUR";
```
Shadouts, line 2: Single quote and double quotes being used inconsistently. I'm a single-quoter for anything that isn't supposed to be readable text for the end user. Can we make it the prettier rule?

src/interaction.js
```diff
@@ -176,7 +176,7 @@ export class KeystoreInteraction {
    * supported.  See the Bowser documentation for more details:
    * https://github.com/lancedikson/bowser
    *
-   * @returns {boolean} whether this interaction is supported
+   * @returns {boolean | Promise<boolean>} whether this interaction is supported
```
Shadouts, line 179: What's going on with this? Returning a union between a `Promise` type and a `boolean` seems like it has a high bug potential since they each are evaluated very differently. A `Promise` is truthy and must be waited for. If this can be async, please make it `Promise<boolean>` only and have non-promise instances `return Promise.resolve(boolean)`

src/ledger.ts
```diff
@@ -308,6 +348,35 @@ export class LedgerBitcoinInteraction extends LedgerInteraction {
...
+    }
+    return this.isV2Supported;
+  }
+
+  /**
+   * Inheriting classes should call the super.run()
+   * as well as set the properties of support before calling their run
+   * in order to check support before calling the actual interaction run
+   */
+  async run() {
```
Shadouts, line 372: Ah, the good ol' "Typescript classes can't use async constructors" obstacle. There was once a day I did it like this, but have since avoided it whenever possible. I find it better to allow for uninitialized class instances while using a `static async` method to return a class instance in a specific state. Something like `static async WithApp(): LedgerInteraction` or `static async WithSupport(version): LedgerInteraction`. Could that work here too? If so, might it make it less messy feeling than requiring a constructor call a method on the parent class? If not, perhaps this needs to be a `protected` method?

Shadouts (approved):
Just some minor stuff and questions.

Shadouts (changes requested):
I accidentally approved in the previous. Meant to request changes (don't know if I could edit my last review).

src/vendor/bcur/index.js
bucko13: It's only inconsistent because we didn't have prettier before. It's being added in files that I'm clicking "save" on for the first time (we should probably add a pre-commit hook tbh). As noted elsewhere, I'm preferring for consistency with other code bases we maintain. If we want to make a change to all our rules to be single quotes, I'm fine with that, but for now inconsistency across projects is a bigger friction point in my opinion.

src/ledger.ts
bucko13: This doesn't return an instance though. `run` methods are meant to be run on instances of the class. This particular method doesn't run anything at all. It will be superseded by inheriting classes which should call `super.run()` to do the `isSupported` check (which throws if that that interaction isn't supported). For example `LedgerExportPublicKey` will be instantiated [here](https://github.com/bucko13/unchained-wallets/blob/8a0047a8e7bba16086de61062e2c4ac2491edb72/src/index.js#L130-L155) and then `run` will be called on it. because `LedgerExportPublicKey#run` calls `super.run()` it will do the isSupported check (i was originally starting to write these manually for each interaction class and it was feeling messy.

src/interaction.js
bucko13: Yep. Good point. the fun challenge of being at the beginning of our TS migration journey

bucko13: So I started by making it only be a Promise. but that had a bunch of a downstream consequences of all the other places that call this (to your point about being handled differently). I opted instead to just rename the `isSupported` in the Ledger interactions that is the only one that needs to be async (because it involves an app interaction to check the version) to be `isAppSupported` and use that for this one particular use case. Makes the change much smaller: [9ddccb2](https://github.com/unchained-capital/unchained-wallets/pull/88/commits/9ddccb2fd7a2b07210178e2091fd8e29d144177f)

src/vendor/bcur/index.js
Shadouts: There is no consistency between code bases. Yellen is using single quotes.

Shadouts (approved):

PR #89, opened 2023-02-15
feat: add wallet policy de/serialization support, wallet registration, address verification

bucko13:
Adds a few new features that leverage wallet policies. This is tooling to let us interact with wallet policies, most importantly to be able to import them to a ledger v2 app/device. 

All of the examples of how to use it can be seen in the tests I added. I'd like to build out more validation, parsing, serialization, etc. for the policies and templates, but this gives a nice minimal viable product that is more than any other JS tools I've seen out there. 

This is the next step required before we can do signing. Once PSBTv2 support is in then we can update the interactions. 

Updates in this PR include:
- [x] A WalletPolicy class for converting to ledger compatible object
- [x] validators for checking policy templates and key origins
- [x] converting from `Braids` (object in unchained-bitcoin) to wallet policies
- [x] A new base interaction class for any interactions that have a prerequisite to register. also accepts a policyHmac to skip registration
- [x] A brand new interaction for verifying address on a ledger device

This can be tested with Caravan's test suite, but I need to push some updates up to Caravan and unchained-bitcoin first.


src/ledger.ts
```diff
@@ -1280,3 +1287,183 @@ export class LedgerSignMessage extends LedgerBitcoinInteraction {
...
+    if (this.policyHmac) return Promise.resolve(this.policyHmac);
+    else {
+      return this.withApp(async (app: AppClient) => {
+        const policy = this.walletPolicy.toLedgerPolicy();
+        const [policyId, policyHmac] = await app.registerWallet(policy);
+        this.policyHmac = policyHmac;
+        this.policyId = policyId;
+        policyHmac;
+      });
+    }
```
tylerchambers, line 1347: redundant else

src/ledger.ts
```diff
@@ -1280,3 +1287,183 @@ export class LedgerSignMessage extends LedgerBitcoinInteraction {
...
+    expected,
+  }: ConfirmAddressConstructorArguments) {
+    super({ braid, name });
+
+    const braidIndex = Number(braid.index);
+    if (braidIndex !== 1 && braidIndex !== 0)
+      throw new Error(`Invalid braid index ${braidIndex}`);
+    this.braidIndex = braidIndex;
+
+    const index = Number(addressIndex);
```
tylerchambers, line 1391: Should already be a number since it's a number in the arguments

src/ledger.ts
```diff
@@ -369,13 +375,14 @@ export class LedgerBitcoinInteraction extends LedgerInteraction {
    * as well as set the properties of support before calling their run
    * in order to check support before calling the actual interaction run
    */
-  async run() {
+  async run(): Promise<unknown> {
```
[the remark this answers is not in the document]
bucko13: Unfortunately not because there are other classes that inherit from it and so if it's void then the children classes that do return something will complain

src/ledger.ts
bucko13: I was unsure how this would be dealt with in non-TS environments so I left it there to just be really really sure.

src/ledger.ts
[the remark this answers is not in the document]
Shadouts: There is a way to reference the polymorphic type to perform conditional typing. Is there is a parent class that's actually an abstract class. Is that `LedgerInteraction`? Then you can do something like 
```
type ChildTypes = LedgerBitcoinInteraction | SomethingElse;

abstract class LedgerInteraction<InstanceType extends ChildTypes = ChildTypes> {
  //  and in here you can reference InstanceType
}
```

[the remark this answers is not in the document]
Shadouts: It probably is too heavy for now, at least until everything is typescriptified or refactored. using `unknown` is like using `any`, it breaks typescripts typing and should be avoided.

src/index.js
```diff
@@ -440,14 +448,64 @@ export function ConfirmMultisigAddress({
         multisig,
         publicKey,
       });
+    case LEDGER: {
+      // TODO: clean this up. The reason for this is that
+      // we're expecting this malleable object `multisig` that
+      // gets passed in but really these interactions should
+      // just get a braid or something derived from it.
```
tylerchambers, line 455: Not sure if it's appropriate here since this will probably be done soon, but I think it would be good if we began logging TODOs in ClickUp and including a ticket ID in the comment.

src/policy.ts
```diff
@@ -0,0 +1,185 @@
...
+
+export type MultisigScriptType = "sh" | "wsh" | "tr";
+
+/**
+ * Takes a Braid instance and translates it into a wallet policy template string
+ * @param {Braid} braid - instance of a braid object from unchained-bitcoin
+ * @returns valid policy template string
+ */
+export const getPolicyTemplateFromBraid = (braid: Braid) => {
+  let scriptType: MultisigScriptType | undefined;
```
tylerchambers, line 49: nit: I don't think we need `| undefined` here since the switch statement covers all possible values.

src/policy.ts
```diff
@@ -0,0 +1,185 @@
...
+      scriptType = "wsh";
+      nested = true;
+      break;
+    default:
+      break;
+  }
+
+  if (!scriptType) {
+    throw new Error(`Unknown address type: ${braid.addressType}`);
+  }
```
tylerchambers, line 72: with undefined removed above typescript's typechecking will prevent the default case here. We could throw in the default.

src/policy.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  const signersString = braid.extendedPublicKeys
+    .map((_, index) => `@${index}/**`)
+    .join(",");
+
+  // TODO: should this always assume sorted?
```
tylerchambers, line 78: imo yes

src/policy.ts
```diff
@@ -0,0 +1,185 @@
...
+      new KeyOrigin({
+        xfp: key.rootFingerprint,
+        xpub: key.base58String,
+        bip32Path: key.path,
+        network: braid.network,
+      })
+  );
+};
+
+export class MutlisigWalletPolicy {
```
tylerchambers, line 97: spelling

src/policy.ts
```diff
@@ -0,0 +1,185 @@
...
+
+export const validateMultisigPolicyScriptType = (template) => {
+  const acceptedScripts = ["sh", "wsh"];
+  let hasMatch = false;
+  for (const script of acceptedScripts) {
+    if (template.match(`^${script}`)) {
+      hasMatch = true;
+      break;
+    }
+  }
```
tylerchambers, line 150: another js FP code golf attempt:

```js
const hasMatch = acceptedScripts.some((script) => template.startsWith(script));`
```

tylerchambers (changes requested):
Looking good. Stylistic suggestions and a spelling mistake

src/policy.ts
bucko13: Yeah. I think I agree but if we want this to be any sort of "reference" for policies then we might have to be more permissive.

src/policy.ts
bucko13: I like it

bucko13: c83d7d9

src/policy.ts
bucko13: Whoops! c83d7d9

src/policy.ts
bucko13: Good call.  c83d7d9

src/policy.ts
bucko13: c83d7d9

.eslintrc.json
```diff
@@ -80,7 +80,10 @@
         "func-names": "error",
         "func-style": [
             "error",
-            "declaration"
+            "declaration",
+            {
+                "allowArrowFunctions": true
```
Shadouts, line 85: :+1:

src/ledger.ts
```diff
@@ -1280,3 +1310,237 @@ export class LedgerSignMessage extends LedgerBitcoinInteraction {
...
+export class LedgerBitcoinV2WithRegistrationInteraction extends LedgerBitcoinInteraction {
+  walletPolicy: MultisigWalletPolicy;
+
+  policyHmac?: Buffer;
+
+  policyId?: Buffer;
+
+  isLegacySupported = false;
+
+  isV2Supported = true;
```
Shadouts, line 1334: Maybe more definition is needed for these for completeness. `readonly`? `const`? `private` with a getter? It's best practice to keep attributes like this private while always providing getters/setters.

src/ledger.ts
```diff
@@ -1280,3 +1310,237 @@ export class LedgerSignMessage extends LedgerBitcoinInteraction {
...
+    });
+  }
+}
+
+interface LedgerRegisterWalletPolicyArguments
+  extends RegistrationConstructorArguments {
+  verify?: boolean;
+}
+export class LedgerRegisterWalletPolicy extends LedgerBitcoinV2WithRegistrationInteraction {
+  verify: boolean;
```
Shadouts, line 1404: Same as above. Also, this is a better example for why to use private attributes but with getters/setters. Should any consumer be allowed to change this at any time or when the class is in any state, or should there be controls? Even if it doesn't need business logic controls now, having a getter/setter while keeping this private will indicate to other developers where the source of truth is controlled. If no gettter/setter, should this be `readonly`?

src/policy.ts
```diff
@@ -0,0 +1,178 @@
...
+import { BitcoinNetwork } from "./types";
+
+export class KeyOrigin {
+  xfp: string;
+
+  bip32Path: string;
+
+  xpub: string;
+
+  network: BitcoinNetwork;
```
Shadouts, line 17: Same as in `ledger.ts`, maybe this needs to use more typescript class features. They would be hard to add later if these get used in unintended ways.

src/policy.ts
```diff
@@ -0,0 +1,178 @@
...
+      })
+  );
+};
+
+export class MultisigWalletPolicy {
+  name: string;
+
+  template: string;
+
+  keyOrigins: KeyOrigin[];
```
Shadouts, line 102: More class stuff.

Shadouts (commented):
Nothing to hold things back, but these comments I've made I think bear some weight. By not putting access modifiers or using getters/setters on classes, attributes can be used in ways which can break state management. It's harder to add them later vs removing them.

src/ledger.ts
bucko13: Apparently this doesn't really work with the complex inheritance that this functionality is built on. I tried it and typescript complains.

bucko13: nvm, it works on the final descendent. So I'm making them readonly on `LedgerBitcoinV2WithRegistrationInteraction` and that settles it nicely.

Shadouts (approved):

PR #91, opened 2023-02-21
Ledger v2 signing interactions

bucko13:
Adds a new signing interaction for ledger v2. 

There is a standalone interaction and there is backwards compatibility built in to the existing interaction such that the new interaction will be used as a fallback.

These changes require the two dependent PRs in unchained-bitcoin to work: [PSBTv2Maps](https://github.com/unchained-capital/unchained-bitcoin/pull/62) and [this patch](https://github.com/Shadouts/unchained-bitcoin/pull/1).

New functionality can be tested out with [these updates](https://github.com/unchained-capital/caravan/pull/287) to the caravan test suite

### DIFF
--- a/src/ledger.ts
+++ b/src/ledger.ts
@@ -913,6 +913,26 @@ export class LedgerExportExtendedPublicKey extends LedgerExportHDNode {
   }
 }
 
+interface LedgerSignMultisigTransactionArguments {
+  network: BitcoinNetwork;
+
+  inputs: TxInput[];
+
+  outputs: object[];
+
+  bip32Paths: string[];
+
+  psbt?: string;
+
+  keyDetails?: KeyDerivation;
+
+  returnSignatureArray?: boolean;
+
+  pubkeys?: Buffer[];
+
+  v2Options?: LedgerV2SignTxConstructorArguments;
+}
+
 /**
  * Returns a signature for a bitcoin transaction with inputs from one
  * or many multisig addresses.
@@ -983,10 +1003,10 @@ export class LedgerSignMultisigTransaction extends LedgerBitcoinInteraction {
    * @param {array<object>} [options.inputs] - inputs for the transaction
    * @param {array<object>} [options.outputs] - outputs for the transaction
    * @param {array<string>} [options.bip32Paths] - BIP32 paths
+   * @param {object} [options.v2Options] - arguments to try with a v2 app
    * @param {string} [options.psbt] - PSBT string encoded in base64
    * @param {object} [options.keyDetails] - Signing Key Details (Fingerprint + bip32 prefix)
    * @param {boolean} [options.returnSignatureArray] - return an array of signatures instead of a signed PSBT (useful for test suite)
-   * @param {object} [options.v2Options] - arguments to try with a v2 app
    */
   constructor({
     network,
@@ -997,7 +1017,7 @@ export class LedgerSignMultisigTransaction extends LedgerBitcoinInteraction {
     keyDetails,
     returnSignatureArray = false,
     v2Options,
-  }) {
+  }: LedgerSignMultisigTransactionArguments) {
     super();
     this.network = network;
     if (!psbt || !keyDetails) {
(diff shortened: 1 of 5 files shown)


src/ledger.ts
```diff
@@ -1538,7 +1557,81 @@ export class LedgerConfirmMultisigAddress extends LedgerBitcoinV2WithRegistratio
...
+
+type InputIndex = number;
+// a Buffer with either a 33-byte compressed pubkey or a 32-byte
+// x-only pubkey whose corresponding secret key was used to sign;
+type PubKey = Buffer;
+// a Buffer with the corresponding signature.
+type SignatureBuffer = Buffer;
+// return type of ledger after signing
+type LedgerSignatures = [InputIndex, PubKey, SignatureBuffer];
+export class LedgerV2SignMultisigTransaction extends LedgerBitcoinV2WithRegistrationInteraction {
```
tylerchambers, line 1583: nit: adding some whitespace here would be good for readability.

src/ledger.ts
```diff
@@ -1538,7 +1557,81 @@ export class LedgerConfirmMultisigAddress extends LedgerBitcoinV2WithRegistratio
...
+    }
+  }
+
+  async signPsbt(): Promise<LedgerSignatures[]> {
+    return this.withApp(async (app: AppClient) => {
+      this.signatures = await app.signPsbt(
+        this.psbt,
+        this.walletPolicy.toLedgerPolicy(),
+        this.policyHmac || null,
+        this.progressCallback
```
tylerchambers, line 1619: is it possible for this to be undefined? is it OK if it is?

src/policy.ts
```diff
@@ -107,7 +107,19 @@ export class MultisigWalletPolicy {
     template: string;
     keyOrigins: KeyOrigin[];
   }) {
-    this.name = name;
+    // this is an unstated restriction from ledger
+    // if we don't check it here then registration will fail
+    // with an opaque error about invalid input data
+    // TODO: should this be left as full length and only
+    // abbreviated when translating to a ledger policy?
```
tylerchambers, line 114: I think truncating here is fine.

tylerchambers (changes requested):
Mostly nits/stylistic suggestions

src/ledger.ts
bucko13: Yeah. Probably won't be used most of the time. The use case (which will be REALLY nice if we can leverage) is for a progress tracker. This gets called anytime an input is signed. But in most cases it will be undefined and not used.

tylerchambers (approved):

src/ledger.ts
```diff
@@ -909,6 +913,26 @@ export class LedgerExportExtendedPublicKey extends LedgerExportHDNode {
   }
 }
 
+interface LedgerSignMultisigTransactionArguments {
+  network: BitcoinNetwork;
+
+  inputs: TxInput[];
+
+  outputs: object[];
```
Shadouts, line 921: I think in typescript `object` is the type of an empty object (`{}`). As you can imagine, it's not a very useful type. Unless this interface will always receive outputs as `{}[]` (read as an array of empty objects), it should be `any[]`.

Leaving it as is will error when you pass it something like `[{ value: true }]` saying something like `Type object has no attribute "value"`

src/policy.ts
```diff
@@ -107,7 +107,19 @@ export class MultisigWalletPolicy {
...
+    if (name.length > 64) {
```
Shadouts, line 115: Maybe keep an eye on this. I saw it reporting name as undefined in the test suite, but things might not be aligning there.

Shadouts (changes requested):

src/policy.ts
bucko13: I think this was just because the fixtures in your environment weren't up to date so an undefined variable was getting passed to the interaction class. If caravan were in typescript too that would've failed during build.

src/ledger.ts
bucko13: I feel like I've seen this work as `object` but i could be wrong. I've updated to be `outputs: Record<string, unknown>[];` in any case until we can get better ts penetration.

bucko13: Seems to work, but doesn't hurt to change to a record in any case. 

Screenshot 2023-02-21 at 6 02 27 PM

Shadouts: Perhaps you are right. I'm looking at reference for it now. I could be confusing `object` with `Object`.

Shadouts (approved):

PR #93, opened 2023-02-22
feat: convert psbt from legacy args

bucko13:
made some updates in the process of testing signing. refactored the api for the ledger interactions to enable it to be a little more consistent and not as tightly coupled to test fixtures. 

I also changed the main entrypoint of the app to typescript and made some other typing improvements. 

CI is failing because requires changes to unchained-bitcoin: https://github.com/unchained-capital/unchained-bitcoin/pull/65.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "compile:watch": "babel --watch --extensions .ts,.js -d lib/ src/ && compile-images",
     "prepublish": "npm run compile",
     "prepare": "npm run compile && husky install",
-    "test": "jest src",
+    "test": "npm run check-types && jest src",
     "testdebug": "node --inspect-brk node_modules/.bin/jest --runInBand lib",
     "docs": "./bin/build-docs.sh",
     "lint": "eslint src --ignore-pattern src/vendor"
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "compile:trezor-dev": "TREZOR_DEV=true babel --extensions .ts,.js --watch -d lib/ src/ && npm run compile-images",
     "compile:watch": "babel --watch --extensions .ts,.js -d lib/ src/ && compile-images",
     "prepublish": "npm run compile",
-    "pretest": "npm run check-types && npm run compile",
     "prepare": "npm run compile && husky install",
     "test": "jest src",
     "testdebug": "node --inspect-brk node_modules/.bin/jest --runInBand lib",
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,7 +40,6 @@ import {
   TrezorConfirmMultisigAddress,
   TrezorSignMessage,
 } from "./trezor";
-import { Braid } from "unchained-bitcoin";
 import { BraidDetails, MultisigWalletConfig } from "./types";
 import { braidDetailsToWalletConfig } from "./policy";
 
@@ -487,7 +486,7 @@ export function ConfirmMultisigAddress({
         expected: multisig.address,
         // this is for the name of the wallet the address being confirmed is from
         name,
-        braidIndex: braidDetails.index,
+        braidIndex: Number(braidDetails.index),
         addressIndex,
         policyHmac,
       });
@@ -587,3 +586,4 @@ export * from "./custom";
 export * from "./hermit";
 export * from "./ledger";
 export * from "./trezor";
+export * from "./policy";
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,6 +41,8 @@ import {
   TrezorSignMessage,
 } from "./trezor";
 import { Braid } from "unchained-bitcoin";
+import { BraidDetails, MultisigWalletConfig } from "./types";
+import { braidDetailsToWalletConfig } from "./policy";
 
 /**
  * Current unchained-wallets version.
@@ -84,7 +86,10 @@ export const INDIRECT_KEYSTORES = {
 export const KEYSTORES = {
   ...DIRECT_KEYSTORES,
   ...INDIRECT_KEYSTORES,
-};
+} as const;
+
+type KEYSTORE_KEYS = keyof typeof KEYSTORES;
+type KEYSTORE_TYPES = (typeof KEYSTORES)[KEYSTORE_KEYS];
 
 /**
  * Return an interaction class for obtaining metadata from the given
@@ -323,8 +328,7 @@ export function SignMultisigTransaction({
   psbt,
   keyDetails,
   returnSignatureArray = false,
-  name,
-  braid,
+  walletConfig,
   policyHmac,
   progressCallback,
 }) {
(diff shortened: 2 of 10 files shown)


bucko13:
It’s occurring to me that maybe a lot of this should be moved to unchained-bitcoin, at least all of the policy code leaving unchained wallets to stay focused on just the interactions.

src/index.ts
```diff
@@ -479,12 +484,14 @@ export function ConfirmMultisigAddress({
...
-      const braid = Braid.fromData(JSON.parse(multisig.braidDetails));
+      const braidDetails: BraidDetails = JSON.parse(multisig.braidDetails);
+      const walletConfig = braidDetailsToWalletConfig(braidDetails);
       return new LedgerConfirmMultisigAddress({
+        ...walletConfig,
         expected: multisig.address,
         // this is for the name of the wallet the address being confirmed is from
         name,
-        braid,
+        braidIndex: Number(braidDetails.index),
```
nk1tz, line 494: Is this clamp for precautionary measure because we don't have control over `braidDetails`? or have to work with pre-existing string values?

bucko13: Yeah, both. There are a lot of non-typescript files that will use this, so this is just a nice fallback. Also there are places where braidDetails comes as a json and then needs to be converted at the point it's passed in as well. I had this break on me without the `Number()` when testing with caravan so this just seemed safer. A lot a whack-a-mole when youre only partially typescript across a family of different libraries/projects.

src/ledger.test.ts
```diff
@@ -6,7 +6,7 @@ import {
   TEST_FIXTURES,
   ROOT_FINGERPRINT,
   TESTNET,
-  Braid,
+  BraidDetails,
```
nk1tz, line 9: I don't think `unchained-bitcoin` exports anything called BraidDetails, let alone a type of this name.

bucko13: it does now! https://github.com/unchained-capital/unchained-bitcoin/pull/64/commits/7c952987d113877006be9880531d3e24d64b3bf6#diff-3ffbdbfbe56d0a02febc16d54bb115634a239ba47d8e9f0c49fb0095ad1f1703

Shadouts: It will.

package.json
```diff
@@ -53,7 +53,6 @@
     "compile:trezor-dev": "TREZOR_DEV=true babel --extensions .ts,.js --watch -d lib/ src/ && npm run compile-images",
     "compile:watch": "babel --watch --extensions .ts,.js -d lib/ src/ && compile-images",
     "prepublish": "npm run compile",
-    "pretest": "npm run check-types && npm run compile",
```
Shadouts, line 56: Typescript type checking is kind of like a test of sorts.  Add check-types to test script?

src/ledger.ts
```diff
@@ -1643,7 +1643,8 @@ export class LedgerV2SignMultisigTransaction extends LedgerBitcoinV2WithRegistra
   async signPsbt(): Promise<LedgerSignatures[]> {
     return this.withApp(async (app: AppClient) => {
       this.signatures = await app.signPsbt(
-        this.psbt,
+        // unchained-bitcoin mimics ledgers' methods
+        this.psbt as unknown as LedgerPsbtV2,
```
nk1tz, line 1647: Don't think you need the double assertion here

src/ledger.test.ts
nk1tz: Ah nvm, the pr we just merged... got it

nk1tz (approved):
🚀

src/ledger.ts
bucko13: You are correct! I was getting an error here at some point but could've been my editor getting screwed up with all the package changes.

src/ledger.test.ts
```diff
@@ -389,12 +390,12 @@ describe("ledger", () => {
     function interactionBuilder(
       policyHmac?: string,
       verify?: boolean,
-      name = "satoshi's wallet",
-      braid = TEST_FIXTURES.braids[0]
+      walletConfig = braidDetailsToWalletConfig(
+        TEST_FIXTURES.braids[0] as unknown as BraidDetails
```
Shadouts, line 394: The double `as` always throws me off, maybe just a preference thing, but I always preferred `<unknown>TEST_FIXTURES.braids[0] as BraidDetails`. If anything, it's a couple characters shorter. I just read it in my head as "Unknown test fixture as BraidDetails".

src/ledger.ts
Shadouts: Here is another place where `<unknown>this.psbt as LedgerPsbtV2` can be used instead of the chained `as`.

Shadouts: I suppose this is working sometimes, but I don't know why. Could it be the cause of the other signature tests in Caravan failing? I believe unchained-bitcoin's `PsbtV2` only provides `serialize` as a common method. Is it possible other calls are being made to different methods which do not exist on `PsbtV2` and that is why the non-P2SH signatures are failing?

If not, this needs to probably have a more clear comment explaining that unchained-bitcoin PsbtV2 really only provides `serialize`.

src/policy.test.ts
```diff
@@ -73,11 +73,13 @@ describe("KeyOrigin", () => {
...
 });
 
-describe("getPolicyTemplateFromBraid", () => {
+describe("getPolicyTemplateFromWalletConfig", () => {
   it("converts braids to valid policies", () => {
     for (const multisig of TEST_FIXTURES.multisigs) {
-      const braid = Braid.fromData(multisig.braidDetails);
-      const template = getPolicyTemplateFromBraid(braid);
+      const walletConfig = braidDetailsToWalletConfig(
+        multisig.braidDetails as unknown as BraidDetails
```
Shadouts, line 80: `<unknown>multisig.braidDetails as BraidDetails`?

Shadouts (approved):
Some comments. The one about the possible PsbtV2 incompatibilities seems concerning to me, but it doesn't have to be a blocker.

package.json
bucko13: Done  6035a7c
I think that was the original idea. Forget why I removed. `npm run compile` might've been messing stuff up.

src/ledger.test.ts
bucko13: turns out wasnt' needed at all, but this is a cool pattern. 
Removed: 6035a7c

src/ledger.ts
bucko13: calls are being made to `copy`. I suppose there might be something different about P2SH PSBTs that allows the two versions to be in parity for those transactions and then break for segwit types.

src/policy.test.ts
bucko13: removed 6035a7c

PR #94, opened 2023-02-22
fix: catch a failed transport.close()

Shadouts:
OS: Linux
Browser: Chrome 110.0.0.0
unchained-wallets: v.0.1.27
Keystore: Ledger

`await transport.close()` fails with error `index.js:1 DOMException: Failed to execute 'releaseInterface' on 'USBDevice': Unable to release interface.` interrupting the completion of many tests. Catching it here (and logging it) allows for tests to complete.


bucko13 (approved):
Thanks!

PR #95, opened 2023-02-24
Wallet config fix

bucko13:
Mostly typescript fixes. Also adds support for `ExportPublicKey` interaction for ledger v2.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,9 +40,13 @@ import {
   TrezorConfirmMultisigAddress,
   TrezorSignMessage,
 } from "./trezor";
-import { MultisigWalletConfig } from "./types";
+import { MultisigWalletConfig, TxInput } from "./types";
 import { braidDetailsToWalletConfig } from "./policy";
-import { unsignedMultisigPSBT, BraidDetails } from "unchained-bitcoin";
+import {
+  unsignedMultisigPSBT,
+  BraidDetails,
+  BitcoinNetwork,
+} from "unchained-bitcoin";
 
 /**
  * Current unchained-wallets version.
@@ -89,7 +93,7 @@ export const KEYSTORES = {
 } as const;
 
 type KEYSTORE_KEYS = keyof typeof KEYSTORES;
-type KEYSTORE_TYPES = (typeof KEYSTORES)[KEYSTORE_KEYS];
+export type KEYSTORE_TYPES = (typeof KEYSTORES)[KEYSTORE_KEYS];
 
 /**
  * Return an interaction class for obtaining metadata from the given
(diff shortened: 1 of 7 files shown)


src/ledger.ts
```diff
@@ -1656,13 +1698,25 @@ export class LedgerV2SignMultisigTransaction extends LedgerBitcoinV2WithRegistra
...
+    return addSignaturesToPSBT(
+      this.network,
+      this.unsignedPsbt,
+      // array of pubkeys as buffers
+      this.signatures.map((sig) => Buffer.from(sig[1])),
+      // array of sigs as buffers
+      this.signatures.map((sig) => Buffer.from(sig[2]))
+    );
+  }
+
```
bucko13, line 1711: @Shadouts so this is how I resolved adding signatures for now (the interactions do need them and the "normal" signing flow wouldn't work without something like this). namely I save the psbt that comes in through the constructor on a `this.unsignedPsbt`. that stays raw and is assumed to v0 since this function is from bitcoinjs-lib and v0 only. Later we can change this to just add the sig directly onto `this.psbt` and run `serialize` on that.

src/ledger.ts
```diff
@@ -809,17 +817,26 @@ export class LedgerExportPublicKey extends LedgerExportHDNode {
...
-      return publicKey;
+        return publicKey;
+      } else {
+        const publicKey = await this.getV2PublicKey();
+        if (this.includeXFP) {
+          let rootFingerprint = await this.getXfp();
+          return { rootFingerprint, publicKey };
+        }
+        return publicKey;
+      }
```
tylerchambers, line 839: the above if returns so we don't need this else

src/index.ts
```diff
@@ -319,6 +323,19 @@ export function ExportExtendedPublicKey({
...
+  inputs?: TxInput[];
+  outputs?: object[];
+  bip32Paths?: string[];
+  psbt: string;
+  keyDetails: { xfp: string; path: string };
+  returnSignatureArray?: boolean;
+  walletConfig: MultisigWalletConfig;
+  policyHmac?: string;
+  progressCallback?: () => void;
+}
```
nk1tz, line 338: Do you have a preference for `interface` over `type`?

They are SO close to being the same thing now, but the main differences as I understand them are:

- `type` has a more succinct syntax (`&` versus `extends` for example)
- `declaration merging` can only be done with `interface` (although I don't like the practice because its a bit "magical")
- `type` can union interfaces together, but not the other way around...

So generally I reach for `type` unless:
- I specifically need to `extend` a class with an interface or have a desire to allow declaration merging (which is never honestly - but maybe if I was creating a library with exported types I **might** want them to be interfaces to allow consumers to extend them via declaration merging??)

nk1tz (approved):
Looks pretty straightforward!

src/index.ts
bucko13: Heh, I've actually leaned in the other direction that basically in most cases interfaces are preferable over types simply because I remember a while ago reading that they're mostly the same but the extending in interfaces is useful so just use them unless you really need types. This seems to be the advice in the [TS docs,](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) although they also say it's basically personal preference unless you really need one feature or the other. The, admittedly simplistic, heuristic I've used is that I like types for more static items, single line declarations, things of that nature, and then interfaces for object and class shapes if that makes sense.

src/ledger.ts
bucko13: done! thanks. [ee17f8b](https://github.com/unchained-capital/unchained-wallets/pull/95/commits/ee17f8b2e99bd3740dbd81f6a5ed0ac8060bf6d3)

src/index.ts
nk1tz: Yeah its not very consequential but `&` and `extends` do the same I think.

Anyways, onto important things! lol

PR #98, opened 2023-03-09
Better support for serializing ledgerPolicyHmacs

bucko13:
- Improves messages for new ledger interactions
- Fixes a bug where transport closing wasn't being awaited 
- changes serialization for better long term support with multiple vendors


nk1tz (approved):
Ok policyHmac being store as a buffer now.

PR #99, opened 2023-03-10
feat(ledger): use bip32path for ledger confirm address api

bucko13:
When working on confirm address on device in Caravan outside of the test suite context, I ran into issues where the API was a little clunky and didn't match with what was going on in the other keystore (trezor). This feels cleaner anyway. Ledger wants braid and address index, which is just available in bip32 path anyway. 

Companion draft PR up in Caravan that uses this functionality: https://github.com/unchained-capital/caravan/pull/296

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -487,8 +487,8 @@ export function ConfirmMultisigAddress({
   multisig,
   publicKey,
   name,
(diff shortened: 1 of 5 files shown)


nk1tz (approved):
👌🏻

PR #101, opened 2023-03-22
replace vendored ledger code with npm package

bucko13:
Also includes a more robust check to see if legacy app is being used for ledger version checking.

I confirmed that this works in Caravan with the two signing tests we know work now (p2sh and p2wsh. nested segwit still 
doesn't work with ledger's new app)

Screenshot 2023-03-22 at 12 06 59 PM
Screenshot 2023-03-22 at 12 08 16 PM


src/ledger.ts
```diff
@@ -232,13 +234,17 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
...
       this.appVersion = response.version;
+      this.appName = response.name;
       return this.appVersion;
     });
   }
 
   async isLegacyApp(): Promise<boolean> {
     const version = await this.setAppVersion();
     const [majorVersion, minorVersion] = version.split(".");
+    // if the name includes "Legacy" than it is legacy app
```
jevidon, line 245: "than" should be "then".

src/ledger.ts
```diff
@@ -232,13 +234,17 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
...
+      this.appName = response.name;
       return this.appVersion;
     });
   }
 
   async isLegacyApp(): Promise<boolean> {
     const version = await this.setAppVersion();
     const [majorVersion, minorVersion] = version.split(".");
+    // if the name includes "Legacy" then it is legacy app
+    if (this.appName && Boolean(this.appName.includes("Legacy"))) return false;
```
tylerchambers, line 246: should this return true?

tylerchambers (changes requested):

src/ledger.ts
bucko13: Ooof. Yes. Not enough coffee for me and not enough storage in my ledger to have both versions of the apps installed apparently

bucko13: Fixed. derp. Thanks for catching

tylerchambers (approved):
I think this is fine now

bucko13:
@tylerchambers I caught up with master and it dismissed your review. Mind giving it another peek?

tylerchambers (approved):
still looks good!

PR #102, opened 2023-03-23
Trim wallet name

bucko13:
ledger will throw an invalid data error if there is trailing whitespace. There's no need to allow this in the policy information anyway so removing when creating the policy for registration and other interactions should avoid this issue entirely.


nk1tz (approved):
Nice and simple

PR #104, opened 2023-03-28
feat(policy): order key origins when instantiating wallet policy

bucko13:
This should resolve an issue of immutability for wallet policy registrations with ledger where if a registration hmac is created with keys in one particular order and then that registration is attempted to be verified against a policy with the keys in a different order, then they won't match and the registration will fail to verify. 

This will be a minor version update since it could break any registrations that were handling the order consistently before passing it to uc-wallets and the order would be changed by the sorting here. 

I tested this in Caravan which you can see in the screenshots below. The first two are registering for the bob key without this change in the uc-wallets dependency and changing the order (which you can see in the key table) 

**First Alice (0) - Bob (1)**
Screenshot 2023-03-27 at 9 46 46 PM

**Then Bob (1) - Alice (0)**
Screenshot 2023-03-27 at 9 44 19 PM

However with this change the registration for bob is the same even if the keys are in a different order.

**Then Bob (1) - Alice (0)**
Screenshot 2023-03-27 at 9 52 31 PM

**Then Bob (1) - Alice (0)**
Screenshot 2023-03-27 at 9 44 19 PM

### DIFF
--- a/src/ledger.ts
+++ b/src/ledger.ts
@@ -1405,8 +1405,6 @@ export abstract class LedgerBitcoinV2WithRegistrationInteraction extends LedgerB
     ...walletConfig
   }: RegistrationConstructorArguments) {
     super();
-    const keyOrigins = getKeyOriginsFromWalletConfig(walletConfig);
-    const template = getPolicyTemplateFromWalletConfig(walletConfig);
     if (policyHmac) {
       const error = validateHex(policyHmac);
       if (error) throw new Error(`Invalid policyHmac`);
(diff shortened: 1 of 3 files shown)


nk1tz (approved):
I'm confused about when the sorting actually happens.

Obviously the line in the constructor of `MultisigWalletPolicy` constructor.
```
this.keyOrigins = keyOrigins.sort((a, b) => a.xpub.localeCompare(b.xpub));
```

But I don't see any calls to `new MultisigWalletPolicy` anymore.

Now in `ledger.ts` we are just calling:
```
this.walletPolicy = MultisigWalletPolicy.FromWalletConfig(walletConfig);
```

Which .... oh then that calls 
```
static FromWalletConfig(config: MultisigWalletConfig): MultisigWalletPolicy {
//,......
    return new this({ name, template, keyOrigins });
  }
```

Got it, `return new this` 🤦🏻

This is why I'm not a fan of classes. Just feels so convoluted to me.

nk1tz (approved):

PR #111, opened 2023-04-06
fix: change old cjs require to esm imports

bucko13:
This was causing some issues in some build contexts. Given the new minimum nodejs requirement, this change should be fine. Confirmed that it compiles without errors as well.


bucko13:
Yeah, I agree on the configs. This is something that will need to be fixed I think at some point but might need some additional refactoring of our modules (imports, exports, etc.) before we can get there properly :-/

PR #116, opened 2023-09-06
Leverage unchained-bitcoin PsbtV2 to track ledger sigs

Shadouts:
This change updates the Ledger Bitcoin v2 interaction to use the unchained-bitcoin PSBTv2 class to track signatures instead of tracking them internal to the interaction class.


bucko13:
Bummer we didn't get this in the migration. Should be moved to https://caravan-bitcoin/caravan.

PR #118, opened 2023-09-14
Trezor: Use browser host for dev URL defaults rather than assume localhost

Shadouts:
This change affects Trezor and changes the default from assuming localhost to assuming whatever the browser host is. Not all development environments are configured to run services at the host machine level. Using browser host as the default will allow for a stateless default which does not have limitations of a localhost configuration. For example: There may be a development environment running remotely from the browser and port mapped to forward the trezor requests to its Trezor Connect dev service.


bucko13 (approved):
🤦

PR #120, opened 2023-09-15
typescript conversions

KevinRiggen:


src/coldcard.ts
Shadouts, line 1: More jsdoc defs to remove in this module. They should be removed and then typescript can be the sole source for api info.

src/coldcard.ts
```diff
@@ -275,11 +280,13 @@ class ColdcardMultisigSettingsFileParser extends ColdcardInteraction {
...
-    let addressType = COLDCARD_BASE_BIP32_PATHS[knownColdcardChroot];
+    let relativePath =
+      knownColdcardChroot &&
+      getRelativeBIP32Path(knownColdcardChroot, this.bip32Path);
+    let addressType;
+    if (knownColdcardChroot !== null) {
+      addressType = COLDCARD_BASE_BIP32_PATHS[knownColdcardChroot];
+    }
     // result could have p2wsh_p2sh or p2sh_p2wsh based on firmware version. Blah!
     if (addressType.includes("_") && !result[addressType.toLowerCase()]) {
```
Shadouts, line 291: I like the additional checking above, but this line will throw if `addressType` is undefined. Give it an initializer on its declaration.

src/coldcard.ts
```diff
@@ -410,15 +405,16 @@ export class ColdcardExportExtendedPublicKey extends ColdcardMultisigSettingsFil
...
 export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
-  /**
-   *
-   * @param {object} options - options argument
-   * @param {string} options.network - bitcoin network
-   * @param {array<object>} options.inputs - inputs for the transaction
-   * @param {array<object>} options.outputs - outputs for the transaction
-   * @param {array<string>} options.bip32Paths - BIP32 paths
-   * @param {object} [options.psbt] - PSBT of the transaction to sign, include it or we will generate it
-   */
```
Shadouts, line 421: I see some jsdoc stuff has been remove, but the ones attached to comments above the definition block also will affect typing.

src/custom.ts
Shadouts, line 1: More jsdoc typedefs which should be removed.

src/hermit.ts
Shadouts, line 1: For jsdoc, this just has `@module` and `@extends` directives, but I wonder if even those might cause a problem. Let's drop them. The goal is to remove jsdoc anyhow.

src/trezor.ts
Shadouts, line 1: More jsdoc stuff.

Shadouts (changes requested):
One logic change issue, but everything else is to do with jsdoc.

Shadouts:
Closing this. It is continued with https://github.com/unchained-capital/unchained-wallets/pull/129

PR #129, opened 2023-10-27
Typescript and v1.0.0!

Shadouts:
Continuing work from https://github.com/unchained-capital/unchained-wallets/pull/120.

This change also upgrades unchained-bitcoin to v1.0.0 and removes jsdocs. Jsdoc type system can conflict with typescript, so type directives were removed from the docstrings.

Many types were broken after converting modules to typescript. A few of these types took some in-depth consideration in how to define their types in order to be typescript compliant.


nk1tz (approved):
This all looks really good.
🥇 mad props.
1.0 let's go!

bucko13:
Conflict in trezor.js

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "unchained-wallets",
-  "version": "0.6.2",
+  "version": "1.0.0",
```
bucko13, line 3: Just leave it to release please. Should be fine with a `feat!` prefix.

package.json
```diff
@@ -37,7 +37,6 @@
     "jest": "^29.4.2",
     "jest-environment-jsdom": "^29.4.1",
     "jest-junit": "^9.0.0",
-    "jsdoc": "^4.0.0",
```
bucko13, line 40: I don't think `npm run docs` will work anymore since it relies on this package. Is there anyway we can keep it but just less feature-ful after the cleanup you did? Otherwise, i'm not against removing it for now. It would be good to revive auto doc generation eventually but doesn't need to block Typescript in my opinion as this feels like more valuable "documentation" anyway. 

If you do remove it you'll also have to remove the script from `bin/` and the link in the README.

src/coldcard.test.ts
```diff
@@ -524,7 +525,7 @@ describe("ColdcardSignMultisigTransaction", () => {
         outputs: transactions[0].outputs,
         bip32Paths: transactions[0].bip32Paths,
       });
-      const result = interaction.request().toBase64();
+      const result = interaction.request().data.toBase64();
```
bucko13, line 528: wow was this just broken?

src/coldcard.ts
```diff
@@ -176,11 +176,8 @@ class ColdcardMultisigSettingsFileParser extends ColdcardInteraction {
    * add a field for rootFingerprint (it can sometimes be calculated
    * if not explicitly included)
    *
-   * @param {Object} file JSON file exported from Coldcard
-   * @returns {Object} the parsed response
-   *
    */
-  parse(file) {
+  parse(file: any) {
```
bucko13, line 180: Based on the code below I think this works as a type declaration for this arg.

```suggestion
  parse(file: Record<string, unknown> | string) {
```

src/coldcard.ts
```diff
@@ -269,31 +267,30 @@ class ColdcardMultisigSettingsFileParser extends ColdcardInteraction {
    * 2. derive deeper if necessary (and able) using functionality
    *    from unchained-bitcoin
    *
-   * @param {Object} result - parsed data from ColdcardJSON
-   * @returns {Object} the desired xpub (if possible)
-   *
    */
-  deriveDeeperXpubIfNecessary(result) {
+  deriveDeeperXpubIfNecessary(result: any) {
```
bucko13, line 271: I think this could be typed at least as a record as well

```suggestion
  deriveDeeperXpubIfNecessary(result: Record<string, unknown>) {
```

src/coldcard.ts
```diff
@@ -410,15 +392,16 @@ export class ColdcardExportExtendedPublicKey extends ColdcardMultisigSettingsFil
...
-   */
+  network: string;
+
+  psbt: any;
+
+  inputs: any[];
+
+  outputs: any[];
+
+  bip32Paths: string[];
```
bucko13, line 403: I think when we can avoid `any` we should. Maybe this caused downstream failures that expanded the scope too much so feel free to reject if you already tried something like this. 

```suggestion
  network: BitcoinNetwork;

  psbt: Record<string, unknown>;

  inputs: Record<string, unknown>[];

  outputs: Record<string, unknown>[];

  bip32Paths: string[];
```

src/coldcard.ts
```diff
@@ -550,6 +529,18 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
  *
  */
 export class ColdcardMultisigWalletConfig {
+  jsonConfig: any;
```
bucko13, line 532: This is already typed actually [where it's called](https://github.com/unchained-capital/unchained-wallets/blob/master/src/index.ts#L580C33-L580C33) in index.ts
```suggestion
  jsonConfig: string | MultisigWalletConfig;
```

src/coldcard.ts
```diff
@@ -550,6 +529,18 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
+
+  name: string;
+
+  requiredSigners: number;
+
+  totalSigners: number;
+
+  addressType: string;
```
bucko13, line 540: This is also available from `unchained-bitcoin`

```suggestion
  addressType: MultisigAddressType;
```

src/coldcard.ts
```diff
@@ -550,6 +529,18 @@ export class ColdcardSignMultisigTransaction extends ColdcardInteraction {
...
+
+  extendedPublicKeys: any[];
```
bucko13, line 542:
```suggestion
  extendedPublicKeys: WalletConfigKeyDerivation[];
```

from [here](https://github.com/unchained-capital/unchained-wallets/blob/master/src/types/index.ts#L20-L24).

src/custom.ts
```diff
@@ -176,14 +168,16 @@ export class CustomExportExtendedPublicKey extends CustomInteraction {
...
-   */
+  network: BitcoinNetwork;
+
+  inputs: any;
+
+  outputs: any;
+
+  bip32Paths: any;
+
+  psbt: any;
```
bucko13, line 179: same here.

src/index.ts
```diff
@@ -489,6 +455,15 @@ export function ConfirmMultisigAddress({
...
   walletConfig,
+}: {
+  keystore: KEYSTORE_TYPES;
+  network: Network;
+  bip32Path: string;
+  multisig: any;
+  publicKey?: string;
+  name?: string;
+  policyHmac?: string;
+  walletConfig?: any;
```
bucko13, line 466:
```suggestion
  walletConfig?: MultisigWalletConfig;
```

src/index.ts
```diff
@@ -489,6 +455,15 @@ export function ConfirmMultisigAddress({
   name,
   policyHmac,
   walletConfig,
+}: {
+  keystore: KEYSTORE_TYPES;
+  network: Network;
+  bip32Path: string;
+  multisig: any;
```
bucko13, line 462: I want to retire this object so bad. It has so many different shapes, it gets mutated, the name is confusing. Just a note for the future since this one has to be `any` for now because of this confusion (or doesn't have to be but it's too complex to address now). 

Small improvement could at least instead be:

```suggestion
  multisig: Record<string, unknown>;
```

since it has to be an object.

src/interaction.test.ts
```diff
@@ -16,16 +16,16 @@ import {
...
 
-    constructor({param}) {
+    constructor({ param }) {
       super();
       this.param = param;
     }
 
     messages() {
-      const messages = super.messages();
+      const messages: any = super.messages();
```
bucko13, line 28: This should be something like this no?

```suggestion
      const messages: Record<string, string>[] = super.messages();
```

src/interaction.ts
```diff
@@ -18,92 +16,82 @@ import { signatureNoSighashType } from "unchained-bitcoin";
...
+ * Represents a message returned by an interaction.
  *
- * @typedef module:interaction.MessageImage
- * @type {Object}
- * @property {string} label - a human-readable label for the image
- * @property {string} mimeType - the MIME type of the image
- * @property {string} data - base64-encoded image data
+ * Message objects may have additional properties.
  */
+type Message = {
```
bucko13, line 63: actually if you've got this type you could export it and use it in the place where I commented on the `any` previously.

src/interaction.ts
```diff
@@ -623,20 +566,13 @@ export class IndirectKeystoreInteraction extends KeystoreInteraction {
    * appropriate kind of `response` object and return the final result
    * of this interaction.
    *
-   * @param {Object} response - the raw response
-   * @returns {Object} the parsed response
-   *
    */
-  parse(response) {
+  parse(response: object) {
```
bucko13, line 570: My understanding is that `Record<string, any>` is preferable here but this is probably fine.

src/ledger.ts
```diff
@@ -180,16 +163,14 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
...
-   * @returns {Promise} does the work of setting up a transport connection
    * @example
    * async run() {
    *   return await this.withTransport(async (transport) => {
    *     return transport.doSomething(); // Not a real Ledger transport API call
    *   });
    * }
    */
-  async withTransport(callback) {
+  async withTransport(callback: (transport: any) => any) {
```
bucko13, line 173: I wonder if we can get the transport type from the ledger library.

src/ledger.ts
```diff
@@ -252,16 +233,14 @@ export class LedgerInteraction extends DirectKeystoreInteraction {
...
-   * @returns {Promise} does the work of setting up an app instance (and transport connection)
    * @example
    * async run() {
    *   return await this.withApp(async (app, transport) => {
    *     return app.doSomething(); // Not a real Ledger bitcoin app API call
    *   });
    * }
    */
-  withApp(callback) {
+  withApp(callback: (app: any, transport: any) => any) {
```
bucko13, line 243: app is definitely available from ledger [here](https://github.com/unchained-capital/unchained-wallets/blob/master/src/ledger.ts#L57). `AppClient`.

src/ledger.ts
```diff
@@ -388,9 +356,10 @@ export abstract class LedgerBitcoinInteraction extends LedgerInteraction {
    * Inheriting classes should call the super.run()
    * as well as set the properties of support before calling their run
    * in order to check support before calling the actual interaction run
-   * @returns {Promise} could return void or some value that the interaction is meant to provide
+   *
+   * The return type has to remain any to get inheritance typing to work.
    */
-  async run(): Promise<unknown> {
+  async run(): Promise<any> {
```
bucko13, line 362: My understanding is that `unknown` is [preferable](https://www.benmvp.com/blog/when-use-typescript-unknown-versus-any/) over `any`

src/ledger.ts
```diff
@@ -1518,7 +1447,7 @@ export class LedgerRegisterWalletPolicy extends LedgerBitcoinV2WithRegistrationI
     return messages;
   }
 
-  async run(): Promise<string> {
```
bucko13, line 1521: Wonder why this was required to be removed.

src/trezor.ts
```diff
@@ -462,25 +428,22 @@ export class TrezorGetMetadata extends TrezorInteraction {
...
-   * @param {string} options.network - bitcoin network
-   * @param {string} bip32Path - the BIP32 path for the HD node
-   * @param {boolean} includeXFP - return xpub with root fingerprint concatenated
-   */
-  constructor({ network, bip32Path, includeXFP }) {
+  bip32Path: string;
+
+  includeXFP: boolean;
+
+  bip32ValidationErrorMessage: any;
```
bucko13, line 444: Would this not need to be a string?

src/trezor.ts
```diff
@@ -708,19 +660,23 @@ export class TrezorExportExtendedPublicKey extends TrezorExportHDNode {
...
-   * @param {object} options - options argument
-   * @param {string} options.network - bitcoin network
-   * @param {UTXO[]} [options.inputs] - inputs for the transaction
-   * @param {TransactionOutput[]} [options.outputs] - outputs for the transaction
-   * @param {string[]} [options.bip32Paths] - BIP32 paths on this device to sign with, one per each input
-   * @param {string} [options.psbt] - PSBT string encoded in base64
-   * @param {object} [options.keyDetails] - Signing Key Details (Fingerprint + bip32 prefix)
-   * @param {boolean} [options.returnSignatureArray] - return an array of signatures instead of a signed PSBT (useful for test suite)
-   */
+  // network: Network
```
bucko13, line 665: delete

src/trezor.ts
```diff
@@ -708,19 +660,23 @@ export class TrezorExportExtendedPublicKey extends TrezorExportHDNode {
...
+  inputs: any[];
+
+  outputs: any[];
+
+  bip32Paths: string[];
+
+  psbt?: string;
+
+  keyDetails: any;
```
bucko13, line 674: I think this can use WalletConfigKeyDerivation for this?

bucko13 (changes requested):
Awesome job. Mostly comments about typing and also a suggestion regarding jsdoc. We can't quite remove that dependency without some other cleanup.

package.json
Shadouts: oh yeah. I had forgotten.

package.json
Shadouts: I removed it to align with the same change made in unchained-bitcoin. The jsdocs comments have been gutted and are now basically just descriptions.

src/coldcard.test.ts
Shadouts: It was broken only due to a change in unchained-bitcoin v1.0.0. I think the psbt is on both the request root and `data`, but it was only typed at the `data` attribute. So, it worked the way it was without typescript. @nk1tz would have more info if he remembers. I think he contributed the typing change on unchained-bitcoin.

src/coldcard.ts
Shadouts: I think these were added as `any` for the exact reason you mention. This was a series of timeboxed changes and the minimum was to get everything into typescript files and building, but not necessarily to fix all the typing. If all the types must be solid, then it might need to wait for another timebox.

src/coldcard.ts
Shadouts: Not so simple. The entire constructor must be reworked if this gets a type. See above about scope.

src/coldcard.ts
Shadouts: All of these constructor initialization values are being pulled from `jsonConfig`. Can the `jsonConfig` actually be definitely typed, or is the entire thing really an `unknown`? Is there any type validation for `jsonConfig`?

src/interaction.test.ts
Shadouts: it's actually a `Message[]` but `Message` is not public. It can be represented as `Record<string, any>[]`, but this is a unit test. I think unit tests should be robust enough to work without typing since they are in the realm of runtime. Setting it as `any` in a world with `unknown` and `any` types will allow a test to express possible returns values which could happen within a codebase which has incomplete type tests.

I'll change it to `unknown`.

src/interaction.ts
Shadouts: I don't think it's correct to export a type just so a test can use it. The test should be emulating a caller against an API. It should only have access to what an outside caller would have.

src/interaction.ts
Shadouts: A few things were extending this. I didn't want to touch it, but I had to to get the others typed for your above request.

package.json
bucko13: Yeah but this dependency is used in a script that is still here. So if we keep the script to generate docs even if it’s just descriptions then we need this dependency I think.

src/ledger.ts
Shadouts: I read the article, and I agree with it, but the article really isn't making the case that it's generally preferred over `any`. It describes an environment with strict type control and which you actually don't know the type of something. I think `unknown` is especially great when dealing with external APIs.

In this case, the type is actually known. `run` inherits an extensive union type. Rather than blowing up the scope to bring in a messy union, I figured we can leave it as `any` for now. Another helpful thing about `any` (for now), is that when we turn on `noExplicitAny` it will show us what we must fix.

Please consider this comment for many other cases where `any` was used instead of `unknown`. In most of them, changing to `unknown` wasn't so simple for this reason and others.

src/coldcard.ts
bucko13: Yeah that’s fair enough. Can I request an update maybe to the top of the readme regarding types and indicate some of these examples I highlight in my comments? It could be a good first task for someone looking to get their feet wet (as we recently had in caravan), so nice to give some explicit direction, also just for keeping track of this not being the state we want long term. 

I didn’t check but are we at least warning for `any` and we can eventually turn to an error?

src/ledger.ts
Shadouts: Inheritance conflict.

src/ledger.ts
bucko13: Yeah so basically `any` is being used to document a type we want to update as a next step. Let’s just mention that in a dedicated comment somewhere like I mentioned elsewhere and then these are all fine with me.

src/trezor.ts
Shadouts: I think this can be removed.

src/ledger.ts
Shadouts: I tried pulling it out of AppClient, but it doesn't seem to be made available.

src/interaction.test.ts
Shadouts: So, for some reason addressing changes have caused typescript error 4053 to start being thrown across the code. I've never encountered this error, but it requires that `Message` be imported to a module just so that class methods can use the inherited return type. I don't know why it wasn't being throw before, and I probably never have encountered it because it seems to be happening due to heavy reliance on inheritance.

So, I have to export `interaction.Message` type now, but this test typing still doesn't work because the `param` field on each object is not part of type `Message` and is only used as sort of a test-id.

src/ledger.ts
Shadouts: Actually, `AppClient` doesn't work. I changed this an there were 5 lines erroring on reference of methods on the class type. For example: `Property 'getWalletPublicKey' does not exist on type 'AppClient'.`

src/trezor.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/trezor.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/ledger.ts
bucko13: That's strange. I wonder if it's because of the old app version or something.

src/trezor.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

package.json
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

package.json
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/coldcard.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/coldcard.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/coldcard.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/coldcard.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/index.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/index.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/interaction.ts
Shadouts: Found to be needed for an error which wasn't previously being thrown. Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/interaction.ts
Shadouts: Changed in 35668a8805582ecc715ddc5e5214ac4de91c22ff

src/ledger.ts
Shadouts: Readme note added in 8a7d8b3ff59d054336ed85b21940f79453a7f606

src/coldcard.ts
Shadouts: Readme note added in 8a7d8b3ff59d054336ed85b21940f79453a7f606

There is no `any` warning. It will error if tsconfig is updated to have `noExplicitAny`.

src/custom.ts
Shadouts: Readme note added in 8a7d8b3ff59d054336ed85b21940f79453a7f606

src/coldcard.ts
bucko13: I think we could do it with an eslint rule:

```
module.exports = {
  "rules": {
    "@typescript-eslint/no-explicit-any": "warn"
  }
};
```

but I don't feel super strongly about it. was just a thought about how to get us to eventually clean these up without blocking this update.

src/interaction.ts
```diff
@@ -18,92 +16,82 @@ import { signatureNoSighashType } from "unchained-bitcoin";
...
- * @type {Object}
- * @property {string} label - a human-readable label for the image
- * @property {string} mimeType - the MIME type of the image
- * @property {string} data - base64-encoded image data
+ * Message objects may have additional properties.
  */
+export type Message = {
+  state?: typeof STATES extends readonly (infer ElementType)[]
+    ? ElementType
+    : never;
```
bucko13, line 66: This was some black magic to me so I had to research a bit. In the off chance that anyone else is reading along: https://blog.logrocket.com/understanding-infer-typescript/

`typeof STATES extends` kind of creates a conditional checking if `STATES` is a readonly array. If it is, infer the type and assign it to `ElementType` and set that as the type, otherwise the type is `never`

(I think that's what's happening. I rewrote to help my own understanding)

src/interaction.ts
```diff
@@ -18,92 +16,82 @@ import { signatureNoSighashType } from "unchained-bitcoin";
...
+  level?: Message["level"];
+  code?: Message["code"] | RegExp;
+  text?: Message["text"] | RegExp;
+  version?: Message["version"] | RegExp;
+};
+
+type FormatType = "buffer" | "hex";
+type FormatReturnType<T extends FormatType> = T extends "buffer"
+  ? Buffer
+  : string;
```
bucko13, line 94: Very cool.

bucko13 (approved):
Really good stuff. Thank you! 🙏

PR #135, opened 2023-11-06
fix: actions yaml bugs

bucko13:
I think we were missing some things, probably because of some out of date/incomplete example readmes. Came up with this solution combining [this](https://github.com/google-github-actions/release-please-action#automating-publication-to-npm) and [this](https://github.com/marketplace/actions/release-please-action#outputs)


sashafklein (approved):
It worked!?!?

Issue #139, opened 2023-12-07
isLegacyApp() check results in false positives


waldenraines:
The check here:  https://github.com/unchained-capital/unchained-wallets/blob/master/src/ledger.ts#L217-L224

Results in false positives as anything with `minorVersion` < 1 will match:

![image](https://github.com/unchained-capital/unchained-wallets/assets/4116405/2a4d3b52-76ca-42f7-afc5-9014ae6add61)

This means that when ledger updates their version to 3.0.0 or later we will be returning `true` from `isLegacyApp()` incorrectly.

jevidon:
The Caravan Library is transitioning to a new home, and as part of this move current issues will be closed and archived. This issue has been resolved as part of the migration - please go check out the new location here: https://github.com/caravan-bitcoin/caravan/issues

PR #140, opened 2023-12-08
Future-proof legacy version checking for Ledger bitcoin app

bigspider:
The current code would consider version 3.0.0 as "legacy".
Also fixing a typo in the comment.


waldenraines:
Fixes https://github.com/unchained-capital/unchained-wallets/issues/139

bucko13:
Applied in monorepo implementation in @caravan/wallets: https://github.com/caravan-bitcoin/caravan/pull/30

PR #141, opened 2024-01-29
fix(trezor): pin connect url to stable version

bucko13:
A recent release of the trezor connect web app ([9.1.10](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/CHANGELOG.md#9110)) introduced a breaking change to trezor interactions. In caravan this can be reproduced by running through a confirm address flow which returns the following:

```
Invalid parameter "bundle/0/multisig/signatures" (= undefined): Required property
```

This change makes sure that, unless told otherwise via env vars, TrezorConnect should use a pinned version of trezor connect web app. This is probably a good idea for the long term anyway since, as this incident indicates, a change to trezor connect can cause a breaking change to any dependency without the ability to opt-in or out. 

This change was tested with a local instance of Caravan and Confirm Address on device worked.

### DIFF
--- a/src/trezor.ts
+++ b/src/trezor.ts
@@ -109,21 +109,18 @@ const TREZOR_DEV =
   env_variables.TREZOR_DEV ||
   env_variables.REACT_APP_TREZOR_DEV ||
   env_variables.VITE_TREZOR_DEV;
+
 try {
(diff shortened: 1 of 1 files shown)


dylan-thompson (approved):
This works!

